Issue #1, opened 2020-07-30
The `list-available` command does not show old versions of operators, only channel heads


joelanford:
Right now, `list-available <package>` can show the versions at each channel head for a particular package, but it doesn't seem possible to know earlier versions.

Ideally, the packagemanifest API would tell me all versions in each channel.

PR #7, opened 2020-08-05
GolangCI-Lint Added with Happy Linter

kensipe:
Signed-off-by: Ken Sipe <kensipe@gmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,25 @@
+linters:
+  auto-fix: false
+  enable:
+    - errcheck
+    - goimports
+    - golint
+    - gosec
+    - misspell
+    - scopelint
+    - unconvert
+    - unparam
+    - interfacer
+    - nakedret
+    - gocyclo
+    - dupl
+    - goconst
+    - lll
+run:
+linters-settings:
+  errcheck:
+    check-type-assertions: true
+  lll:
+    line-length: 250
+  dupl:
+    threshold: 400
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all build
+.PHONY: all build lint
 all: build
 
 build:
@@ -6,3 +6,12 @@ build:
 
 install: build
 	install bin/kubectl-operator $(shell go env GOPATH)/bin
+
+GOLANGCI_LINT_VER = "1.29.0"
+lint:
+#	scripts/golangci-lint-check.sh
+ifneq (${GOLANGCI_LINT_VER}, "$(shell ./bin/golangci-lint --version 2>/dev/null | cut -b 27-32)")
+	@echo "golangci-lint missing or not version '${GOLANGCI_LINT_VER}', downloading..."
+	curl -sSfL "https://raw.githubusercontent.com/golangci/golangci-lint/v${GOLANGCI_LINT_VER}/install.sh" | sh -s -- -b ./bin "v${GOLANGCI_LINT_VER}"
+endif
+	./bin/golangci-lint --timeout 3m run
--- a/internal/pkg/action/operator_install.go
+++ b/internal/pkg/action/operator_install.go
@@ -137,9 +137,10 @@ func (i *OperatorInstall) getPackageChannel(pm *operatorsv1.PackageManifest) (*o
 		i.Channel = pm.Status.DefaultChannel
 	}
 	var packageChannel *operatorsv1.PackageChannel
-	for idx, ch := range pm.Status.Channels {
+	for _, ch := range pm.Status.Channels {
+		ch := ch
 		if ch.Name == i.Channel {
-			packageChannel = &pm.Status.Channels[idx]
+			packageChannel = &ch
 		}
 	}
 	if packageChannel == nil {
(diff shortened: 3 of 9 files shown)


kensipe:
I tool 1 approach for fixing "using-reference-to-loop-iterator-variable" of not ignoring the index and using the index into the collection... but it could have be solved with `s := s` which is more and more common.  Let me know if you prefer that.

kensipe:
It's worth adding 
```
golangci-lint version
golangci-lint has version 1.29.0
```
I'm happy to add details to the Makefile + scripts here or on another PR

joelanford:
Awesome! Yeah let's add a `lint` target and a step in the github actions workflows.

kensipe:
roger that

.golangci.yml
```diff
@@ -0,0 +1,25 @@
...
+    - gocyclo
+    - dupl
+    - goconst
+    - lll
+run:
+linters-settings:
+  errcheck:
+    check-type-assertions: true
+  lll:
+    line-length: 250
```
joelanford, line 23: We recently outright removed `lll` in the SDK linting. But 250 seems like a completely reasonable limit to catch something egregious.

internal/pkg/action/operator_install.go
```diff
@@ -51,7 +51,10 @@ func (i *OperatorInstall) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&i.CreateOperatorGroup, "create-operator-group", "C", false, "create operator group if necessary")
 
 	fs.VarP(&i.InstallMode, "install-mode", "i", "install mode")
-	fs.MarkHidden("install-mode")
+	err := fs.MarkHidden("install-mode")
+	if err != nil {
+		log.Print(`requested flag "install-mode" missing`)
```
joelanford, line 56: `panic` here maybe? This error would only happen if we screw up and change the name of the flag without changing the call to `MarkHidden`.

internal/pkg/action/operator_uninstall.go
```diff
@@ -48,9 +48,9 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 	}
 
 	var sub *v1alpha1.Subscription
-	for _, s := range subs.Items {
+	for i, s := range subs.Items {
 		if u.Package == s.Spec.Package {
-			sub = &s
+			sub = &subs.Items[i]
```
joelanford, line 53: Does this also solve the problem?
```suggestion
	for i := range subs.Items {
		s := &subs.Items[i]
		if u.Package == s.Spec.Package {
			sub = s	
```

If so, I think I prefer this style better so that we don't mix `subs.Items[i]` and `s` in the same loop, which could be confusing.

kensipe: it does... but we could also skip the index `i` completely with 

```
for _ , s := range subs.Items {
    s := s
		if u.Package == s.Spec.Package {
			sub = &s
```

the main diff is `s := s` copies a value into a new mem location and then is deref with `&s` and `&subs.Items[i]` will deref the value from the original mem address.   (barring compiler / runtime optimizations)

joelanford: Yeah let's do that then. Is this kind of problem common? Or would be make sense to add a comment about this where we have to do it, so someone doesn't come along and think "that looks like a no-op... let's remove this line!"

kensipe: it's very common!    It always shows up for data-driven go tests at a min..  example: https://github.com/kudobuilder/kudo/blob/9db03a9f2be3d7bfe457ebc000881ad4433a27d4/pkg/kudoctl/cmd/diagnostics/print_test.go#L192
also entry #1 of the common mistakes wiki: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

Makefile
```diff
@@ -1,8 +1,12 @@
...
 
 build:
 	go build -o bin/kubectl-operator
 
 install: build
 	install bin/kubectl-operator $(shell go env GOPATH)/bin
+
+lint:
+	scripts/golangci-lint-check.sh
+	golangci-lint --timeout 3m run
```
joelanford, line 12: Instead of adding a separate script, wdyt about putting the script into a `golangci-lint` make target kind of like [this](https://github.com/operator-framework/operator-sdk/blob/f04ad0730a655ec0f3aee06a573e3b74205b44af/internal/plugins/helm/v1/scaffolds/internal/templates/makefile.go#L111-L121), except that it would check presence/version in `./bin`, download the right version of `golangci-lint` if necessary into `./bin`, and then always uses it from there.

Developers might be working on different projects that download tools in `$GOBIN`, so this approach would avoid different projects fighting with each other if they use different versions of the same tool.

kensipe: I can switch to that... I would prefer to run from our bin actually.
I try to stay away from script functions in my make file however... it is less portable... among other things... but this makes sense for this.   I will have it tomorrow.

kensipe:
@joelanford  rebased and ready to go

PR #23, opened 2020-10-05
Refactor for cli and pkg/action decoupling and deduplication

joelanford:


varshaprasad96 (commented):
Tested this PR. Everything works well, except for the [issue](https://github.com/operator-framework/kubectl-operator/issues/24) while uninstalling the operator. However, this refactor doesn't seem to be the problem as it occurs in master too.

Issue #24, opened 2020-11-03
Error while uninstalling operator


varshaprasad96:
Tried installing `cockroachdb` operator using the commands as given in `demo.sh` with the binary built from master.

Steps followed:
1. Install operator - `kubectl-operator install cockroachdb --create-operator-group -v 2.1.1`
Output - Operator is installed successfully.

2. Uninstall operator - `kubectl-operator uninstall cockroachdb --delete-crds --delete-operator-groups`
Output:

```
subscription "cockroachdb" deleted
uninstall operator: delete clusterserviceversion "": resource name may not be empty
```

When I check for the csv available in cluster after running the uninstall command, I get this:
```
➜  kubectl-operator git:(main) k get csv --all-namespaces | grep -i cockroachdb
default           cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
kube-node-lease   cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
kube-public       cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
kube-system       cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
olm               cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
operators         cockroachdb.v2.1.1   CockroachDB      2.1.1     cockroachdb.v2.0.9   Succeeded
```

The crds are not deleted either:
```
➜  kubectl-operator git:(main) k get crds --all-namespaces | grep -i cockroachdb
cockroachdbs.charts.helm.k8s.io               2020-11-03T03:01:41Z
```

Expected output: The operator is uninstalled successfully and all the resources are deleted.

joelanford:
Thanks @varshaprasad96! What version of OLM are you using? I'm not able to reproduce this with OLM v0.16.1

joelanford:
And _now_ I just ran into this. So it seems to be intermittent or triggered by something a little more specific.

PR #26, opened 2020-11-04
Fix: Set obj name and namespace during uninstall

varshaprasad96:
Set name and namespace for objects obtained from unmarshalling install plan while uninstalling the operator.

Closes: #24


internal/pkg/action/operator_uninstall.go
```diff
@@ -157,6 +158,7 @@ func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, install
 			Version: step.Resource.Version,
 			Kind:    step.Resource.Kind,
 		})
+		obj.SetName(step.Resource.Name)
```
joelanford, line 161: :+1:

But I don't think we need to unmarshal the `step.Resource.Manifest` anymore. We can just instantiate `obj` and then set GVK and Name since that's all we need to delete the object.
```go
obj := &unstructured.Unstructured{}
obj.SetGroupVersionKind(schema.GroupVersionKind{
	Group:   step.Resource.Group,
	Version: step.Resource.Version,
	Kind:    step.Resource.Kind,
})
obj.SetName(step.Resource.Name)
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -165,6 +167,13 @@ func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, install
 		if step.Resource.Kind == "ServiceAccount" && obj.GetNamespace() == "" {
 			obj.SetNamespace(installPlanKey.Namespace)
 		}
```
joelanford, line 169: This section can be removed since it will be covered by the added `bundle.IsSupported()` block.

joelanford (changes requested):

joelanford (approved):
/lgtm

PR #27, opened 2020-11-05
fix(uninstall): rely on OLM to delete operator resources via CSV deletion

varshaprasad96:
When multiple install plans are present for single operator,
select the resources of install plan containing the installed CSV
while uninstalling the operator.

### DIFF
--- a/internal/pkg/action/operator_uninstall.go
+++ b/internal/pkg/action/operator_uninstall.go
@@ -7,7 +7,6 @@ import (
 
 	v1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
 	"github.com/spf13/pflag"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -66,21 +65,12 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		return fmt.Errorf("operator package %q not found", u.Package)
 	}
 
-	// Since the install plan is owned by the subscription, we need to
-	// read all of the resource references from the install plan before
-	// deleting the subscription.
-	var crds, csvs, others []controllerutil.Object
-	if sub.Status.InstallPlanRef != nil {
-		ipKey := types.NamespacedName{
-			Namespace: sub.Status.InstallPlanRef.Namespace,
-			Name:      sub.Status.InstallPlanRef.Name,
-		}
-		var err error
-		crds, csvs, others, err = u.getInstallPlanResources(ctx, ipKey)
-		if err != nil {
-			return fmt.Errorf("get install plan resources: %v", err)
-		}
+	csv, err := u.getInstalledCSV(ctx, sub)
+	if err != nil {
+		return fmt.Errorf("get installed CSV %q: %v", sub.Status.InstalledCSV, err)
 	}
+
+	crds := getCRDs(csv)
 
 	// Delete the subscription first, so that no further installs or upgrades
 	// of the operator occur while we're cleaning up.
@@ -96,9 +86,10 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		}
 	}
 
-	// Delete CSVs and all other objects created by the install plan.
-	objects := append(csvs, others...)
-	if err := u.deleteObjects(ctx, objects...); err != nil {
+	// OLM puts an ownerref on every namespaced resource to the CSV,
+	// and an owner label on every cluster scoped resource. When CSV is deleted
+	// kube and olm gc will remove all the referenced resources.
+	if err := u.deleteObjects(ctx, csv); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


internal/pkg/action/operator_uninstall.go
```diff
@@ -140,46 +136,54 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
...
 }
 
-func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, installPlanKey types.NamespacedName) (crds, csvs, others []controllerutil.Object, err error) {
-	installPlan := &v1alpha1.InstallPlan{}
-	if err := u.config.Client.Get(ctx, installPlanKey, installPlan); err != nil {
-		return nil, nil, nil, fmt.Errorf("get install plan: %v", err)
+func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, ipNamespace, installedCSV string) (crds, csvs, others []controllerutil.Object, err error) {
+	installPlanList := &v1alpha1.InstallPlanList{}
+	if err := u.config.Client.List(ctx, installPlanList, client.InNamespace(ipNamespace)); err != nil {
+		return nil, nil, nil, errors.New("error getting install plan resources")
```
jmrodri, line 142: any reason not to just do `fmt.Errorf("error getting install plan resources")` to be consistent with the rest of the code.

internal/pkg/action/operator_uninstall.go
```diff
@@ -190,3 +194,14 @@ func contains(haystack []string, needle string) bool {
...
 	return false
 }
+
+func isIPWithInstalledCSV(ip v1alpha1.InstallPlan, installedCSV string) bool {
+	for _, step := range ip.Status.Plan {
+		if step.Resource.Kind == csvKind {
+			if step.Resource.Name == installedCSV {
+				return true
+			}
+		}
```
jmrodri, line 204: Seems like this could be an `&&`:

```golang
if step.Resource.King == csvKind && step.Resource.Name == installedCSV {
    return true
}
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -2,6 +2,7 @@ package action
 
 import (
 	"context"
+	"errors"
```
jmrodri, line 5: I don't see the need for this new import.

jmrodri (changes requested):
2 small issues

internal/pkg/action/operator_uninstall.go
```diff
@@ -71,12 +70,8 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 	// deleting the subscription.
 	var crds, csvs, others []controllerutil.Object
 	if sub.Status.InstallPlanRef != nil {
```
joelanford, line 72: I don't think we need to pay attention to `sub.Status.InstallPlanRef` anymore, since we don't use it to figure out the install plan key.

My instinct here is to locate the install plan separately from getting its resources, and change this `if` statement to check whether we found an install plan.

So something like this:
```go
ip, err := u.getInstallPlanForCSV(sub.Status.StartingCSV)
if err != nil { ... }

var crds, csvs, others []controllerutil.Object
if ip != nil {
        var err error
        crds, csvs, others = u.getInstallPlanResources(*ip)  
}
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -190,3 +193,12 @@ func contains(haystack []string, needle string) bool {
...
 }
+
+func isIPWithInstalledCSV(ip v1alpha1.InstallPlan, installedCSV string) bool {
+	for _, step := range ip.Status.Plan {
+		if step.Resource.Kind == csvKind && step.Resource.Name == installedCSV {
+			return true
+		}
+	}
+	return false
+}
```
joelanford, line 204: For now, I think we should check the install plan's `spec.ClusterServiceVersionNames`, and select an install plan only if it is length 1 and is equal to the CSV name we're looking for.

joelanford: And maybe return an error if we find our csv name, but others are listed.

varshaprasad96: Have modified the implementation. By `others are listed`, do you mean when we find an installplan which has multiple `ClusterServiceNames` - among which is the installed CSV, we return an error because we may have dependent resources specified in other csvs (maybe of another operator) ?

joelanford: Yep, exactly!

internal/pkg/action/operator_uninstall.go
```diff
@@ -140,15 +137,23 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
...
 }
 
-func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, installPlanKey types.NamespacedName) (crds, csvs, others []controllerutil.Object, err error) {
-	installPlan := &v1alpha1.InstallPlan{}
-	if err := u.config.Client.Get(ctx, installPlanKey, installPlan); err != nil {
-		return nil, nil, nil, fmt.Errorf("get install plan: %v", err)
+func (u *OperatorUninstall) getInstallPlanForCSV(ctx context.Context, installedCSV, ipNamespace string) (*v1alpha1.InstallPlan, error) {
+	installPlanList := &v1alpha1.InstallPlanList{}
+	if err := u.config.Client.List(ctx, installPlanList, client.InNamespace(ipNamespace)); err != nil {
+		return nil, fmt.Errorf("error getting install plan resources from namespace %s", ipNamespace)
```
joelanford, line 143: Nit just to match existing error formatting:
```suggestion
		return nil, fmt.Errorf("list install plans in namespace %q: %v", ipNamespace, err)
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -140,15 +137,23 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
 	return waitForDeletion(ctx, u.config.Client, objs...)
 }
 
-func (u *OperatorUninstall) getInstallPlanResources(ctx context.Context, installPlanKey types.NamespacedName) (crds, csvs, others []controllerutil.Object, err error) {
-	installPlan := &v1alpha1.InstallPlan{}
-	if err := u.config.Client.Get(ctx, installPlanKey, installPlan); err != nil {
-		return nil, nil, nil, fmt.Errorf("get install plan: %v", err)
+func (u *OperatorUninstall) getInstallPlanForCSV(ctx context.Context, installedCSV, ipNamespace string) (*v1alpha1.InstallPlan, error) {
```
joelanford, line 140: Nit, we know the subscription namespace is the same as `u.config.Namespace`, so I think we can simplify the signature and use `u.config.Namespace` anywhere we'd use `ipNamespace`.

```suggestion
func (u *OperatorUninstall) getInstallPlanForCSV(ctx context.Context, installedCSV) (*v1alpha1.InstallPlan, error) {
```

internal/pkg/action/operator_uninstall.go
varshaprasad96: Got it! Will modify it.

internal/pkg/action/operator_uninstall.go
```diff
@@ -140,15 +137,28 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
...
+		return nil, fmt.Errorf("list install plans in namespace %q: %v", u.config.Namespace, err)
 	}
 
-	for _, step := range installPlan.Status.Plan {
-		obj := &unstructured.Unstructured{}
+	for _, ip := range installPlanList.Items {
+		csvNames := ip.Spec.ClusterServiceVersionNames
+		if !containsInstalledCSV(installedCSV, csvNames) {
+			continue
+		}
```
joelanford, line 150: Can we reuse the existing `contains()` function here?

internal/pkg/action/operator_uninstall.go
```diff
@@ -140,15 +137,28 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
...
+		if len(csvNames) == 1 {
+			return &ip, nil
+		}
+		return nil, fmt.Errorf("deleting %s can also delete other dependent resources", installedCSV)
```
joelanford, line 154:
```suggestion
		// TODO: use alternate method to ensure only resources from
		//       startingCSV are uninstalled.
		return nil, fmt.Errorf("cannot uninstall %q resources from multi-CSV install plan", installedCSV)
```

joelanford:
Based on some discussion with the OLM team, I think we can immediately improve on solving this problem by ignoring the install plan altogether and deleting only the CSV and CRDs.

We can lookup the CSV using `Name: sub.Status.InstalledCSV, Namespace: sub.GetNamespace()`.

And then lookup the CRDs from `csv.Status.RequirementStatus` filtering on `req.Kind == crdKind`.

internal/pkg/action/operator_uninstall.go
```diff
@@ -140,43 +133,29 @@ func (u *OperatorUninstall) deleteObjects(ctx context.Context, objs ...controlle
...
-		return nil, nil, nil, fmt.Errorf("get install plan: %v", err)
+func (u *OperatorUninstall) getObjectsToDelete(ctx context.Context, key types.NamespacedName) (csvs, crds []controllerutil.Object, err error) {
+	installedCSV := &v1alpha1.ClusterServiceVersion{}
+	if err = u.config.Client.Get(ctx, key, installedCSV); err != nil {
+		return
 	}
 
-	for _, step := range installPlan.Status.Plan {
-		obj := &unstructured.Unstructured{}
+	csvs = append(csvs, installedCSV)
```
joelanford, line 142: Two nits:
1. Now that we're guaranteed to have only one CSV, it might be cleaner to put the "get CSV" functionality into a separate `func (u *OperatorUninstall) getInstalledCSV(ctx, subscription)` function, and call that directly from `Run()`. Then we'd pass just the one csv to `deleteObjects()`.
2. We can make this function responsible for just returning the CRDs from the CSV, so it could be `func getCRDs(csv)`

And one suggestion:
1. When we get the CSV, we need to explicitly set the GVK with  `csv.SetGroupVersionKind()` so that the log output when deleting the CSV doesn't have an empty value for the kind being deleted.

internal/pkg/action/operator_uninstall.go
```diff
@@ -66,22 +65,13 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
-			Name:      sub.Status.InstallPlanRef.Name,
-		}
-		var err error
-		crds, csvs, others, err = u.getInstallPlanResources(ctx, ipKey)
-		if err != nil {
-			return fmt.Errorf("get install plan resources: %v", err)
-		}
+	csv, err := u.getInstalledCSV(ctx, sub)
+	if err != nil {
+		return fmt.Errorf("installed csv not found %s", sub.Status.InstalledCSV)
```
joelanford, line 70:
```suggestion
		return fmt.Errorf("get installed CSV %q: %v", sub.Status.InstalledCSV, err)
```

joelanford (approved):
/lgtm

PR #29, opened 2020-12-14
Improve uninstall command

varshaprasad96:
Refactor uninstall command to clean up all the resources,
and report whether package existed correctly.

Co-authored by: Eric Stroczynski <ericstroczynski@gmail.com>

### DIFF
--- a/internal/pkg/action/operator_uninstall.go
+++ b/internal/pkg/action/operator_uninstall.go
@@ -34,6 +34,14 @@ func NewOperatorUninstall(cfg *Configuration) *OperatorUninstall {
 	}
 }
 
+type ErrPackageNotFound struct {
+	PackageName string
+}
+
+func (e ErrPackageNotFound) Error() string {
+	return fmt.Sprintf("package %q not found", e.PackageName)
+}
+
 func (u *OperatorUninstall) Run(ctx context.Context) error {
 	if u.DeleteAll {
 		u.DeleteCRDs = true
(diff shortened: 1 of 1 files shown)


internal/pkg/action/operator_uninstall.go
```diff
@@ -107,6 +140,11 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		}
 	}
 
+	// If csvObj is nil and crds is not present, it means the package was not found
+	if subObj == nil && len(crds) == 0 {
```
estroz, line 144:
```suggestion
	// If no objects were cleaned up, it means the package was not found
	if subObj == nil && csvObj == nil && len(crds) == 0 {
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -57,16 +65,41 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+
+		// This value can be empty which will cause errors.
+		if csvKey.Name != "" {
+			csv := &v1alpha1.ClusterServiceVersion{}
+			if err := u.config.Client.Get(ctx, csvKey, csv); err != nil && !apierrors.IsNotFound(err) {
+				return fmt.Errorf("error getting installed CSV %q: %v", csvKey.Name, err)
+			} else if err == nil {
+				crds = getCRDs(csv)
+			}
+			csvObj = csv
```
estroz, line 90: (optional) you could move all of this into `OperatorUninstall.getInstalledCSV()`.

estroz (changes requested):
One addition to "no package found" condition then LGTM

estroz (approved):
/lgtm

internal/pkg/action/operator_uninstall.go
```diff
@@ -107,6 +140,11 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		}
 	}
 
+	// If no objects were cleaned up, it means the package was not found
+	if subObj == nil && csvObj == nil && len(crds) == 0 {
+		return &ErrPackageNotFound{u.Package}
```
joelanford, line 145: Since we know `subObj` is never nil by this point (right?), I think we can completely remove this conditional block and use `ErrPackageNotFound` in line 65.

internal/pkg/action/operator_uninstall.go
```diff
@@ -81,7 +114,7 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 	// OLM puts an ownerref on every namespaced resource to the CSV,
 	// and an owner label on every cluster scoped resource. When CSV is deleted
 	// kube and olm gc will remove all the referenced resources.
-	if err := u.deleteObjects(ctx, csv); err != nil {
+	if err := u.deleteObjects(ctx, csvObj); err != nil {
```
joelanford, line 117: If we come out of the CSV fetching with a `nil` csvObj, that means we will have a `nil` CSV and no CRDs to delete (since we find the CRDs from the CSV), right?

Do we need to check to see if the CSV is nil before trying to delete it here?

If it is `nil`, I think that means we'll only be deleting the subscription and possibly the operator group and NOT erroring out on a missing CSV? Seems reasonable since its possible there's a problem with the subscription that prevented the CSV from being created.

joelanford (commented):
Apologies for the late review, but a few more comments and suggestions.

Since this is merged already, I don't mind taking a stab at any other tweaks that come from my comments.

internal/pkg/action/operator_uninstall.go
```diff
@@ -57,16 +65,41 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		return fmt.Errorf("operator package %q not found", u.Package)
 	}
 
-	csv, err := u.getInstalledCSV(ctx, sub)
-	if err != nil {
-		return fmt.Errorf("get installed CSV %q: %v", sub.Status.InstalledCSV, err)
+	var subObj, csvObj controllerutil.Object
+	var crds []controllerutil.Object
+	if sub != nil {
```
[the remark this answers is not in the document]
estroz: You're correct, this condition can be removed due to the above check. This PR is a port of https://github.com/operator-framework/operator-sdk/pull/4303 so there might be some incongruities that should be resolved. `subObj` was necessary due to some strange nil comparison behavior I experienced when writing the linked PR, and made the code a little cleaner. It might not make sense here given your comments below.

internal/pkg/action/operator_uninstall.go
estroz: A nil check should be performed in `deleteObjects()` like in https://github.com/operator-framework/operator-sdk/pull/4303.

joelanford:
I finally got around to submitting PR #31 that makes these changes based on my suggestions.

PR #37, opened 2021-04-05
kubectl operator list-operands library code

exdx:
library code for listing custom resources on-cluster that can then be fed to a command or other callers


internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, c := range crds {
+		// setup GVR
+		gvr := schema.GroupVersionResource{
+			Group:    c.Name,
+			Version:  c.Version,
+			Resource: strings.ToLower(c.Kind) + "s", // is this always a valid assumption?
+		}
+
+		// TODO use namespaced MetaClient if CR is namespace-scoped
+		r, err := o.config.MetaClient.Resource(gvr).List(ctx, metav1.ListOptions{})
```
joelanford, line 102: Instead of using a client-go MetaClient, is it possible to use controller-runtime client's `List` function with `unstructured.UnstructuredList` or `metav1.PartialObjectMetadataList`?

See [`List` implementation in `controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime/blob/11dfabf5e2e83cdb8cba28fedcda0ae5e10474f0/pkg/client/client.go#L215-L225).

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		},
+	}
+	bindOperatorListCustomResourcesFlags(cmd.Flags(), CRLister)
+	return cmd
+}
+
+func bindOperatorListCustomResourcesFlags(fs *pflag.FlagSet, CRLister *action.OperatorListCustomResources) {
+	fs.BoolVarP(&CRLister.ListAll, "list-all", "y", true, "List all CRs in the result without batching. Defaults to true.")
+	fs.StringVarP(&CRLister.Namespace, "namespace", "n", "default", "Set namespace in which to look for operator.")
```
joelanford, line 70: The `--namespace` flag is configured globally and setup as part of the `config`, so no need to add it to the command.

I _would_ recommend adding the `-A, --all-namespaces` flag to this subcommand (see `kubectl operator list`'s implemetation for an example) so that command line users can optionally view CRs from all namespaces rather than just the namespace setup in their kubeconfig context.

internal/cmd/root.go
```diff
@@ -42,6 +42,7 @@ operators from the installed catalogs.`,
 		newOperatorListAvailableCmd(&cfg),
 		newOperatorDescribeCmd(&cfg),
 		newVersionCmd(),
+		newOperatorListCustomResourcesCmd(&cfg),
```
joelanford, line 45: Move this up to group it with the other operator subcommands. How about between ListAvailable and Describe?

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,71 @@
...
+			if len(args) == 0 {
+				log.Fatal("did not provide operator or package name")
+			}
+			CRLister.PackageName = args[0]
+
+			op, err := CRLister.FindOperator(cmd.Context())
+			if err != nil {
+				log.Fatal(err)
+			}
+			fmt.Printf("found operator %s", op.Name)
```
joelanford, line 29: `kubectl` doesn't write log messages like this unless verbosity is set very high. Since `kubectl operator` doesn't have verbosity (at least not yet), I recommend removing these print statements.

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type CustomResourceLister interface {
+	List(ctx context.Context, ownedCRDS []v1alpha1.CRDDescription) (metav1.PartialObjectMetadataList, error)
+}
+
+type CustomResourceDefinitionFinder interface {
+	FindOperator(ctx context.Context, packageName, namespace string) (*v2alpha1.Operator, error)
+	Unzip(ctx context.Context, operator *v2alpha1.Operator) ([]v1alpha1.CRDDescription, error)
+}
```
joelanford, line 25: What's the purpose of these interfaces?

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	config  *Configuration
+	ListAll bool
+	PackageName string
+	Namespace string
+}
+
+func NewOperatorListCustomResources(cfg *Configuration) *OperatorListCustomResources {
+	return &OperatorListCustomResources{
+		config: cfg,
+		Namespace: cfg.Namespace,
```
joelanford, line 38: Any reason to make `Namespace` its own struct field? Or can we always grab it from the config?

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+// FindOperator finds an operator object on-cluster provided a package and namespace. Returns a copy of the operator.
+func (o *OperatorListCustomResources) FindOperator(ctx context.Context) (v2alpha1.Operator, error) {
```
joelanford, line 43: Other existing commands (e.g. `upgrade`, `uninstall`) list subscriptions and find the one that contains the specified package. This `FindOperator` function could be a good candidate for reuse in all of the commands that take an operator name as input.

Could you open an issue that covers moving this to a central place (maybe `pkg/operator`?) and refactoring commands that take operator names as input to use this?

joelanford, line 43: Functions that return (<someStruct>, error) typically return pointers for the struct so that they can return `nil, err` to ambiguity about validity of the struct when an error is returned. 
 
```suggestion
func (o *OperatorListCustomResources) FindOperator(ctx context.Context) (*v2alpha1.Operator, error) {
```

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	opKey := types.NamespacedName{
+		Name:      o.PackageName,
+		Namespace: o.Namespace,
+	}
+	operator := v2alpha1.Operator{}
+	err := o.config.Client.Get(ctx, opKey, &operator)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return v2alpha1.Operator{}, fmt.Errorf("could not find operator for package %s in namespace %s", o.PackageName, o.Namespace)
```
joelanford, line 52: There's an error type that exists for this case. Perhaps we can reuse?
```suggestion
			return nil, ErrPackageNotFound{o.PackageName}
```

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		return v2alpha1.Operator{}, err
```
joelanford, line 54:
```suggestion
		return nil, err
```

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	if operator.Status.Components == nil {
+		return nil, fmt.Errorf("could find underlying components for operator %s", operator.Name)
+	}
+	for _, resource := range operator.Status.Components.Refs {
+		if resource.Kind == v1alpha1.ClusterServiceVersionKind {
+			csvKey.Name = resource.Name
+			csvKey.Namespace = resource.Namespace
+			break
+		}
+		return nil, fmt.Errorf("could find underlying CSV for operator %s", operator.Name)
```
joelanford, line 74: Is the first `Ref` always expected to be the CSV? If not, this will return an error, when we should probably keep iterating.

Perhaps initialize `csvKey` as `var csvKey *types.NamespacedName`. Then set it to `&types.NamespacedName{resource.Namespace, resource.Name}` when we find the CSV. Then after the loop, check that `csvKey` is not nil?

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	err := o.config.Client.Get(ctx, csvKey, &csv)
+	if err != nil {
+		return nil, fmt.Errorf("could not get %s CSV on cluster: %s", csvKey.String(), err)
+	}
+
+	// check if owned CRDs are defined on the csv first
+	if reflect.DeepEqual(csv.Spec.CustomResourceDefinitions, v1alpha1.CustomResourceDefinitions{}) || len(csv.Spec.CustomResourceDefinitions.Owned) == 0 {
```
joelanford, line 83: Looks like `csv.Spec.CustomResourceDefinitions` is a concrete struct value, so probably no need to check that its empty since it can never be `nil`. Therefore, you can directly check `len(csv.Spec.CustomResourceDefinitions.Owned) == 0` without the first check.

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, fmt.Errorf("no owned CustomResourceDefinitions specified on CSV %s, no custom resources to display", csvKey.String())
+	}
+
+	return csv.Spec.CustomResourceDefinitions.Owned, nil
+}
+
+// List takes in a list of CRDs and finds the associated CRs metadata on-cluster using the client-go meta client.
+// Note: List() can return a potentially unbounded list that callers may need to paginate.
+func (o *OperatorListCustomResources) List(ctx context.Context, crds []v1alpha1.CRDDescription) (result metav1.PartialObjectMetadataList, resErr error) {
```
joelanford, line 92: Any reason for the named return values? I typically try to avoid using them unless there's a specific need because it makes `return` statements harder to grok.

This problem is also mentioned in the Tour of Go: https://tour.golang.org/basics/7

joelanford (commented):
Looks like a great start! I went through and did a first pass review, mainly with an eye toward consistency with `kubectl` user expectations and consistency with patterns used elsewhere in `kubectl-operator`.

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,71 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
+	"github.com/operator-framework/kubectl-operator/internal/pkg/action"
```
joelanford, line 7: Nit: can you separate the `github.com/operator-framework/kubectl-operator` imports into a third group?

internal/pkg/action/config.go
```diff
@@ -2,9 +2,11 @@ package action
 
 import (
 	"context"
+	"k8s.io/client-go/metadata"
```
joelanford, line 5: Nit: non-local, non-stdlib imports go in a separate (second group)

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/operators/v2alpha1"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
```
joelanford, line 15: You can group all of these together since they're all non-local.

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,71 @@
...
+						_, err := fmt.Scanf("%s", &cont)
+						if err != nil {
+							log.Fatal(err)
+						}
+						if cont != "y" {
+							break OUTPUT
+						}
+					}
+				}
+			}
```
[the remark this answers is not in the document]
exdx: ok that makes sense, by default the output will be nicely formatted and if they are looking for the json output (to pipe to another command for example) they can use a flag

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,71 @@
...
+			crds, err := CRLister.Unzip(cmd.Context(), &op)
+			if err != nil {
+				log.Fatal(err)
+			}
+			fmt.Printf("found owned crds #%v", crds)
+
+			list, err := CRLister.List(cmd.Context(), crds)
+			if err != nil {
+				log.Fatal(err)
+			}
```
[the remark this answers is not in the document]
exdx: Right, part of the issue is that these additional functions are also potentially to be used as library functions for other projects (like the console team that wants to implement similar functionality on its backend by vendoring in this code). So I'm writing both a CLI command and library code. Maybe I can refactor the additional functions out to a `/lib` or something to make it more clear. I agree with the pattern you suggest in this context.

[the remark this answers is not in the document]
joelanford: As far as I know the console team would need to string these three functions together in this same way right? If so, I still think a single exported entrypoint that can be re-used between this CLI and the console team is the way to go.

internal/pkg/action/operator_list_custom_resources.go
exdx: ohh I see what you mean -- this is a bug, thanks for picking it up

internal/pkg/action/operator_list_custom_resources.go
exdx: I generally also tend not to use named return values, although in this case I thought it could be helpful to have a named `Results` return variable. Totally fine with me to refactor back to standard return values.

joelanford: Yeah I figured that was the reason. I used to do that with functions that just appended to the return value in a loop, but I've tripped myself up often enough on naked returns that I just put `var out []ReturnType` as the first line in those functions and call it a day.

The only time I think I really use named return values is when I need a `defer`-ed function to modify the values on the way out.

internal/cmd/operator_list_custom_resources.go
exdx: My thinking with the `--namespace` flag was that if I am a user in one namespace, I can query for the operator object in another namespace and find the CRs associated with it. I will remove this flag, but this limits the command to only CRs for operators in the current namespace context and across the entire cluster. I think this sounds fine.

internal/pkg/action/operator_list_custom_resources.go
exdx: I wrote them to help me organize the function signatures. Once I pull everything out into the library we can revisit, maybe we don't need them

internal/cmd/operator_list_custom_resources.go
exdx: switching to using the `log` library like in other commands

internal/cmd/operator_list_custom_resources.go
joelanford: You misunderstood my command. You'll get `--namespace` for free. You can access it via `config.Namespace`. If `--namespace` is unset, it will be the value of the namespace from the context. If `--namespace` is set, it will be whatever it is set to.

After having talked about this more today, I think we explicitly _don't_ want an `--all-namespaces` flag, because we're always looking for a specific operator in a specific namespace.

joelanford:
May as well bump `operator-framework/api` to the latest and greatest. Looks like the one we're currently on is pretty old.

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,35 @@
...
+	PackageName string
+}
+
+func NewOperatorListCustomResources(cfg *Configuration) *OperatorListCustomResources {
+	return &OperatorListCustomResources{
+		config: cfg,
+	}
+}
+
+func (o *OperatorListCustomResources) Run(ctx context.Context) (*unstructured.UnstructuredList, error) {
```
joelanford, line 24: As far as I know, this is the only function console needs. Nothing else should be exported, at least until we know there's a specific reason to make it public.

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,35 @@
...
+	opKey := types.NamespacedName{
+		Name:      o.PackageName,
+		Namespace: o.config.Namespace,
+	}
+
+	result, err := lib.ListAll(ctx, o.config.Client, opKey)
```
joelanford, line 30: The intent of `pkg/action` is to provide the main Go-library style entrypoints for the high level functions provided by kubectl-operator's CLI, such that other clients (e.g. console) can use them whole-cloth without having to think about the internals.

That generally means the only exported API in this package should be the structs used by the CLI and their `Run` functions. Anything else should be the exception rather than the rule.

At a minimum, helper functions like this should be put in separate packages more aligned with their concern. If they're only relevant for one particular action (like this one is), I think we should leave it right here in this file and unexport it.

If and when we see that there is code duplication or opportunity for re-use, then that's what I would say we should move functions to a helper package.

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	lib "github.com/operator-framework/kubectl-operator/pkg/action"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// OperatorListCustomResources knows how to find and list custom resources given a package name and namespace.
+type OperatorListCustomResources struct {
```
joelanford, line 13:
```suggestion
type OperatorListOperands struct {
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/operators/v2alpha1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
joelanford, line 16: I would suggest:
1. Combining ` internal/pkg/action/operator_list_custom_resources.go` and this file.
2. Putting the combined file at `pkg/action/operator_list_operands.go` so that it can be imported for external client's uses.
3. Unexporting everything except `OperatorListOperands` and its `Run` function.
4. Moving `internal/pkg/action/config.go` to `pkg/action/config.go`

internal/pkg/action/operator_list_custom_resources.go
```diff
@@ -0,0 +1,35 @@
...
+	config      *Configuration
+	PackageName string
+}
+
+func NewOperatorListCustomResources(cfg *Configuration) *OperatorListCustomResources {
+	return &OperatorListCustomResources{
```
joelanford, line 19:
```suggestion
func NewOperatorListOperands(cfg *Configuration) *OperatorListOperands {
	return &OperatorListOperands{
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,138 @@
...
+			return nil, fmt.Errorf("package %s not found", key.Name)
+		}
+		return nil, err
+	}
+	return &operator, nil
+}
+
+// Unzip finds the CSV referenced by the provided operator and then inspects the spec.customresourcedefinitions.owned
+// section of the CSV to return a list of APIs that are owned by the CSV.
+func Unzip(ctx context.Context, client client.Client, operator *v2alpha1.Operator) ([]v1alpha1.CRDDescription, error) {
```
joelanford, line 33: For now, I'd suggest making `FindOperator`, `Unzip`, `List`, and `ListAll` unexported functions on the `OperatorListOperands` struct.

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,138 @@
...
+	if operator.Status.Components == nil {
+		return nil, fmt.Errorf("could not find underlying components for operator %s", operator.Name)
+	}
+	for _, resource := range operator.Status.Components.Refs {
+		if resource.Kind == v1alpha1.ClusterServiceVersionKind {
+			csvKey.Name = resource.Name
+			csvKey.Namespace = resource.Namespace
+			break
+		}
+	}
```
joelanford, line 46: Is it safe to assume that the operator status refs will always include references to the owned crds? If so, can we skip the CSV business and grab the CRDs directly from the operator status?

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,138 @@
...
+
+	crdDescs, err := Unzip(ctx, client, operator)
+	if err != nil {
+		return nil, err
+	}
+
+	var result unstructured.UnstructuredList
+	for _, crd := range crdDescs {
+		list, err := List(ctx, client, crd, opKey.Namespace)
```
joelanford, line 129: I think we want to search all the target namspaces as defined by the operator group in `opKey.Namespace`, right?

If so, my suggestion would be to:
1. List operator groups (make sure there's exactly 1)
2. Extract `status.Namespaces` from the operator group
3. List operands across the cluster
4. Filter the results from 3 based on the operator group's target namespaces.

If you do it this way, I don't think you'll have to worry about CRD scope since you can just always include items where `item.GetNamespace() == ""`

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,76 @@
+package cmd
```
joelanford, line 1: Maybe just remove this file for now since it is not critical for feature freeze?

I have opinions, but I don't want to block this PR getting in :)

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,76 @@
...
+				}
+				_ = tw.Flush()
+			} else if *output == "json" {
+				for _, cr := range list.Items {
+					j, _ := cr.MarshalJSON()
+					fmt.Printf("#%v", string(j))
+				}
+			} else if *output == "yaml" {
+				//TODO
+			}
```
joelanford, line 58: You can combine the validity check and the decision about which output format to use if you define a variable for the write function and the set that variable based on the value of output. Something like this:
```go
var writeOutput func(io.Writer, *unstructured.UnstructuredList) error
switch output {
case "json":
	writeOutput = writeJSON
case "yaml":
	writeOutput = writeYAML
case "":
	writeOutput = writeTable
default:
	log.Fatal("invalid value for flag output %q, expected one of %s", output, strings.Join(validOutputs, "|"))
}
```

And you can do this towards the top of the `Run` function before calling `crLister.Run`

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+	}
+
+	bindOperatorListCustomResourcesFlags(cmd.Flags(), output)
```
joelanford, line 62: Nit: no real need for a string pointer and a separate function in my mind:
```go
output := ""
validOutputs := []string{"json", "yaml"}
cmd := &cobra.Command{ ... }
cmd.Flags().StringVarP(&output, "output", "o", output, fmt.Sprintf("Output format. One of: %s", strings.Join(validOutputs, "|"))
```

internal/cmd/operator_list_custom_resources.go
```diff
@@ -0,0 +1,76 @@
...
+func newOperatorListCustomResourcesCmd(cfg *action.Configuration) *cobra.Command {
+	crLister := action.NewOperatorListCustomResources(cfg)
+	output := new(string)
+
+	cmd := &cobra.Command{
+		Use:     "list-custom-resources <operator>",
+		Aliases: []string{"list-crs"},
+		Short:   "List custom resources for an operator",
+		Long:    "List all custom resources on-cluster associated with a particular operator",
+		Args:    cobra.MaximumNArgs(1),
```
joelanford, line 26: We always expect exactly one argument for the operator name right?
```suggestion
		Args:    cobra.ExactArgs(1),
```

If you make this change, you don't have to check `len(args) == 0` in the `Run` function.

joelanford (commented):
My big concern is making the public API surface only as big as necessary. This PR is the first time we'll be making _any_ of the Go code in this repo public, so I think we should be judicious.

So most of this review is my thoughts on doing that and giving other clients what they would need to list operands for a particular operator.

pkg/action/custom_resource_lister.go
exdx: ok,done

exdx: pulling out the `Configuration` to its standalone file is producing a pretty big diff so right now I'm just going to import the existing struct and we can pull it out in another PR or commit

joelanford: Sounds good. We just have to do that before external projects can use it because `action.Configuration` is needed when constructing the action struct.

pkg/action/custom_resource_lister.go
exdx: They could yes, based on a label selector that OLM applies when installing the CRDs from the bundle. But the CRDs could also be on-cluster without OLM installing them, in which case they would not be on the operator. I'm not sure.

internal/cmd/operator_list_custom_resources.go
exdx: ok

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,165 @@
...
+
+func NewOperatorListOperands(cfg *internal.Configuration) *OperatorListOperands {
+	return &OperatorListOperands{
+		config: cfg,
+	}
+}
+
+func (o *OperatorListOperands) Run(ctx context.Context) (*unstructured.UnstructuredList, error) {
+	opKey := types.NamespacedName{
+		Name:      o.PackageName,
```
joelanford, line 33: Don't operator names include the namespace?
```suggestion
		Name:      fmt.Sprintf("%s.%s", o.PackageName, o.config.Namespace),
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,165 @@
...
+		Namespace: o.config.Namespace,
+	}
```
joelanford, line 35: Operator is cluster-scoped and includes the namespace in its name, right? 
```suggestion
	opKey := types.NamespacedName{
		Name:      fmt.Sprintf("%s.%s", o.PackageName, o.config.Namespace),
	}
```

internal/cmd/operator_list_custom_resources.go
exdx: will revisit these review comments in a follow-up PR when we add the cmd back :)

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+	// query for CRs in these namespaces
+	ogList := v1.OperatorGroupList{}
+	options := client.ListOptions{Namespace: o.config.Namespace}
+	if err := o.config.Client.List(ctx, &ogList, &options); err != nil {
+		return nil, err
+	}
+	if len(ogList.Items) != 1 {
+		return nil, fmt.Errorf("unexpected number (%d) of operator groups found in namespace %s", len(ogList.Items), o.config.Namespace)
+	}
+	namespaces := ogList.Items[0].Spec.TargetNamespaces
```
joelanford, line 148: To account for namespaces matched using `ogList.Items[0].Spec.Selector`:
```suggestion
	namespaces := ogList.Items[0].Status.Namespaces
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// List takes in a CRD description and finds the associated CRs on-cluster.
+// List can return a potentially unbounded list that callers may need to paginate.
+func (o *OperatorListOperands) list(ctx context.Context, crdDesc v1alpha1.CRDDescription, namespaces []string) (*unstructured.UnstructuredList, error) {
+	result := &unstructured.UnstructuredList{}
+
+	list := unstructured.UnstructuredList{}
+	gvk := schema.GroupVersionKind{
+		Group:   crdDesc.Name,
```
joelanford, line 104: We need the Group here. Not the name. Looking at a CSV, it looks like `crdDesc.Name` contains the actual name of the CRD object. I don't see a Group field in `crdDesc` so we may need to `client.Get()` each CRD.

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+	options := client.ListOptions{Namespace: o.config.Namespace}
+	if err := o.config.Client.List(ctx, &ogList, &options); err != nil {
+		return nil, err
+	}
+	if len(ogList.Items) != 1 {
+		return nil, fmt.Errorf("unexpected number (%d) of operator groups found in namespace %s", len(ogList.Items), o.config.Namespace)
+	}
+	namespaces := ogList.Items[0].Spec.TargetNamespaces
+
+	var result unstructured.UnstructuredList
```
joelanford, line 150: Need to set GVK for `v1/List` to make this valid:
```suggestion
	var result unstructured.UnstructuredList
	result.SetGroupVersionKind(schema.GroupVersionKind{
		Version: "v1",
		Kind:    "List",
	})
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+		Version: crdDesc.Version,
+		Kind:    crdDesc.Kind,
+	}
+	result.SetGroupVersionKind(gvk)
+	if err := o.config.Client.List(ctx, &list); err != nil {
+		return nil, err
+	}
+
+	// trim down CRs in list to match target namespaces
+	if namespaces == nil {
```
joelanford, line 114: This will account for `nil` or empty.
```suggestion
	if len(namespaces) == 0 {
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+		for _, cr := range list.Items {
+			result.Items = append(result.Items, cr)
+		}
+	}
+	return &result, nil
+}
+
+func inNamespace(ns string, namespaces []string) bool {
+	for _, n := range namespaces {
+		if n == ns {
```
joelanford, line 165: When watching all namespaces, a targetNamespace will be the empty string.
```suggestion
		if n == ns || n == "" {
```

pkg/action/custom_resource_lister.go
```diff
@@ -0,0 +1,170 @@
...
+func (o *OperatorListOperands) list(ctx context.Context, crdDesc v1alpha1.CRDDescription, namespaces []string) (*unstructured.UnstructuredList, error) {
+	result := &unstructured.UnstructuredList{}
+
+	list := unstructured.UnstructuredList{}
+	gvk := schema.GroupVersionKind{
+		Group:   crdDesc.Name,
+		Version: crdDesc.Version,
+		Kind:    crdDesc.Kind,
+	}
+	result.SetGroupVersionKind(gvk)
```
joelanford, line 108:
```suggestion
	list.SetGroupVersionKind(gvk)
```

pkg/action/custom_resource_lister.go
exdx: I will update this, but I've never seen the `status.namespaces` used versus targetnamespaces. I assume based on our conversation the status field is the one we want, but we can always change

pkg/action/custom_resource_lister.go
exdx: good catch

pkg/action/custom_resource_lister.go
exdx: ah bummer... ok we will have to GET the CRD to find the group

pkg/action/operator_list_operands.go
```diff
@@ -0,0 +1,183 @@
...
+	result := &unstructured.UnstructuredList{}
+
+	// get crd group from crd name
+	crd := apiextensionsv1.CustomResourceDefinition{}
+	crdKey := types.NamespacedName{
+		Name: crdDesc.Name,
+	}
+	err := o.config.Client.Get(ctx, crdKey, &crd)
+	if err != nil {
+		return nil, nil
```
joelanford, line 109: Need to return an error here if one occurs.
```suggestion
		return nil, fmt.Errorf("get crd %q: %v", crdKey.String(), err)
```

PR #38, opened 2021-04-07
Add kubectl operator list-operands command

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

### DIFF
--- a/internal/cmd/operator_list_operands.go
+++ b/internal/cmd/operator_list_operands.go
@@ -1,0 +1,140 @@
+package cmd
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/duration"
+	"sigs.k8s.io/yaml"
+
+	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+func newOperatorListOperandsCmd(cfg *action.Configuration) *cobra.Command {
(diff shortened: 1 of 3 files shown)


internal/cmd/operator_list_operands.go
```diff
@@ -0,0 +1,128 @@
...
+
+func newOperatorListOperandsCmd(cfg *action.Configuration) *cobra.Command {
+	l := action.NewOperatorListOperands(cfg)
+	output := ""
+	validOutputs := []string{"json", "yaml"}
+
+	cmd := &cobra.Command{
+		Use:   "list-operands <operator>",
+		Short: "List operands of an installed operator",
+		Args:  cobra.ExactArgs(1),
```
exdx, line 31: nit: could we have a `Long` section that describes some of the caveats with this command as well? 

For example, if your operands are cluster-scoped, they will be listed even if you specify a namespace. If your operator is backed by a CSV in a bad state, the list-operands call will fail since the list could be misleading. In lieu of any docs on this command the help string would be great for users.

internal/cmd/operator_list_operands.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := tw.Flush(); err != nil {
+		return err
+	}
+	if _, err := w.Write(buf.Bytes()); err != nil {
+		return err
+	}
+	return nil
+}
+
+func writeJSON(w io.Writer, operands *unstructured.UnstructuredList) error {
```
exdx, line 99: Does this JSON formatting leaving the output in a state where another tool like `kubectl` can process it? One question I have is whether `kubectl` can take a JSON list as input and delete every member. I was envisioning that the `-o json` option would enumerate every entry in the list as a separate row so that it can easily be piped to other commands, but if a list works, then that's great as well

exdx (commented):
Nice!

internal/cmd/operator_list_operands.go
joelanford: Yep! This list works and is how kubectl typically outputs multiple objects.

exdx:
LGTM

PR #43, opened 2021-04-14
Create and set timeouts globally via a flag

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -1,6 +1,11 @@
 package cmd
 
 import (
+	"context"
+	"reflect"
+	"time"
+	"unsafe"
+
 	"github.com/spf13/cobra"
 
 	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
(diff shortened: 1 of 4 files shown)


internal/cmd/root.go
```diff
@@ -24,14 +29,30 @@ operators available for installation, and install and uninstall
...
+	var (
+		cfg     action.Configuration
+		timeout time.Duration
+		cancel  context.CancelFunc
+	)
 
-	var cfg action.Configuration
+	flags := cmd.PersistentFlags()
 	cfg.BindFlags(flags)
+	flags.DurationVar(&timeout, "timeout", 1*time.Minute, "The amount of time to wait before giving up on an operation.")
```
exdx, line 40: I like how by default this is equal to the same amount of time we now retry installplans. So there's some consistency in OLM behavior across projects. cc @benluddy

internal/cmd/root.go
```diff
@@ -47,3 +68,10 @@ operators from the installed catalogs.`,
 
 	return cmd
 }
+
+func setContext(cmd *cobra.Command, ctx context.Context) { //nolint:golint
```
exdx, line 72: Looks funny, but this is the only way to set the unexported field I guess. 

One thing worth noting is mentioned in https://github.com/spf13/cobra/pull/1118#discussion_r562959128 where this root command context would be propagated down to all children. This is what we want via this PR obviously, and if the user wanted to change it on a per-invocation basis they can set the `--timeout` flag on the CLI to whatever. This change would also encourage to build smaller command surfaces that can be used independently of others. 

so a `kubectl operator <command> --timeout 5m | kubectl operator <command> --timeout 10m` could be appropriate in certain cases

exdx (commented):
/lgtm

PR #45, opened 2021-04-19
feat: operator uninstall with operands

### LINKED ISSUE #39: [RFE] Delete operator and operands safely in one command
With the new `kubectl operator list-operands` command we have the ability to obtain a list of Custom Resources associated with an operator. This list can then be fed to `kubectl delete` to delete the CRs on-cluster. 

For example, `kubectl operator list-operands <operator> -o json | jq $filter | kubectl delete -f -`

This RFE is for wrapping operand AND operator deletion in one command while ensuring its safe. Some design considerations:

- There already exists an `kubectl operator uninstall` command that deletes objects on cluster. This command can also delete CRDs, subscriptions, and other resources. We could simply modify this command to meet expected behavior. 
- Another option is a new `kubectl operator delete` command with a narrower scope than `uninstall`, but that would duplicate logic and potentially confuse users
- Any deletion command would need to handle multi-tenancy and dependency concerns (e.g. only allow deleting CRDs if no other operator owns them, don't allow deleting an operator that another operator depends on, etc.)
- The command should probably have a warning and display the list of CRs to be removed before deleting them, with an additional manual opt in (with a `--force` override)

exdx:
Signed-off-by: Daniel Sover <dsover@redhat.com>

Closes #39 

Removes the subscription, optionally the operatorgroup, and the CSV and operands (if the CSV is found on cluster). The `Operator` object associated with the CSV will remain on-cluster if there are any CRDs still on-cluster after `uninstall` runs, since it does not remove CRDs currently.

### DIFF
--- a/internal/cmd/operator_uninstall.go
+++ b/internal/cmd/operator_uninstall.go
@@ -15,8 +15,29 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "uninstall <operator>",
-		Short: "Uninstall an operator",
-		Args:  cobra.ExactArgs(1),
+		Short: "Uninstall an operator and operands",
+		Long: `Uninstall removes the subscription, operator and optionally operands managed by the operator as well as 
+the relevant operatorgroup.
+
+Warning: this command permanently deletes objects from the cluster. Running uninstall concurrently with other operations
+could result in undefined behavior. 
+
+The uninstall command first checks to find the subscription associated with the operator. It then 
+lists all operands found throughout the cluster for the operator
+specified if one is found. Since the scope of an operator is restricted by
+its operator group, this search will include namespace-scoped operands from the
+operator group's target namespaces and all cluster-scoped operands. 
+
+The operand-deletion strategy is then considered if any operands are found on-cluster. One of cancel|ignore|delete. 
+By default, the strategy is "cancel", which means that if any operands are found when deleting the operator abort the 
+uninstall without deleting anything. 
+The "ignore" strategy keeps the operands on cluster and deletes the subscription and the operator.
+The "delete" strategy deletes the subscription, operands, and after they have finished finalizing, the operator itself.
+
+Setting --delete-operator-groups to true will delete the operatorgroup in the provided namespace if no other active 
+subscriptions are currently in that namespace, after removing the operator. The subscription and operatorgroup will be 
+removed even if the operator is not found.`,
+		Args: cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			u.Package = args[0]
 			if err := u.Run(cmd.Context()); err != nil {
@@ -30,8 +51,7 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
 }
 
 func bindOperatorUninstallFlags(fs *pflag.FlagSet, u *internalaction.OperatorUninstall) {
-	fs.BoolVarP(&u.DeleteAll, "delete-all", "X", false, "enable all delete flags")
-	fs.BoolVar(&u.DeleteCRDs, "delete-crds", false, "delete all owned CRDs and all CRs")
 	fs.BoolVar(&u.DeleteOperatorGroups, "delete-operator-groups", false, "delete operator groups if no other operators remain")
 	fs.StringSliceVar(&u.DeleteOperatorGroupNames, "delete-operator-group-names", nil, "specific operator group names to delete (only effective with --delete-operator-groups)")
+	fs.VarP(&u.OperandStrategy, "operand-strategy", "s", "determines how to handle operands when deleting the operator, one of cancel|ignore|delete")
 }
--- a/internal/pkg/action/operator_uninstall.go
+++ b/internal/pkg/action/operator_uninstall.go
@@ -9,10 +9,10 @@ import (
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/operator-framework/kubectl-operator/internal/pkg/operand"
 	"github.com/operator-framework/kubectl-operator/pkg/action"
 )
 
@@ -20,12 +20,10 @@ type OperatorUninstall struct {
 	config *action.Configuration
 
 	Package                  string
-	DeleteAll                bool
-	DeleteCRDs               bool
+	OperandStrategy          operand.DeletionStrategy
 	DeleteOperatorGroups     bool
 	DeleteOperatorGroupNames []string
-
-	Logf func(string, ...interface{})
+	Logf                     func(string, ...interface{})
 }
 
 func NewOperatorUninstall(cfg *action.Configuration) *OperatorUninstall {
@@ -44,11 +42,6 @@ func (e ErrPackageNotFound) Error() string {
 }
 
 func (u *OperatorUninstall) Run(ctx context.Context) error {
-	if u.DeleteAll {
-		u.DeleteCRDs = true
-		u.DeleteOperatorGroups = true
-	}
-
 	subs := v1alpha1.SubscriptionList{}
 	if err := u.config.Client.List(ctx, &subs, client.InNamespace(u.config.Namespace)); err != nil {
 		return fmt.Errorf("list subscriptions: %v", err)
@@ -69,62 +62,56 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 	csv, csvName, err := u.getSubscriptionCSV(ctx, sub)
 	if err != nil && !apierrors.IsNotFound(err) {
 		if csvName == "" {
-			return fmt.Errorf("get subscription CSV: %v", err)
+			return fmt.Errorf("get subscription csv: %v", err)
 		}
-		return fmt.Errorf("get subscription CSV %q: %v", csvName, err)
+		return fmt.Errorf("get subscription csv %q: %v", csvName, err)
 	}
 
-	// Deletion order:
-	//
-	// 1. Subscription to prevent further installs or upgrades of the operator while cleaning up.
-	// 2. CustomResourceDefinitions so the operator has a chance to handle CRs that have finalizers.
-	// 3. ClusterServiceVersion. OLM puts an ownerref on every namespaced resource to the CSV,
-	//    and an owner label on every cluster scoped resource so they get gc'd on deletion.
+	// find operands related to the operator on cluster
+	lister := action.NewOperatorListOperands(u.config)
+	operands, err := lister.Run(ctx, u.Package)
+	if err != nil {
+		return fmt.Errorf("list operands for operator %q: %v", u.Package, err)
+	}
+	// validate the provided deletion strategy before proceeding to deletion
+	if err := u.validStrategy(operands); err != nil {
+		return fmt.Errorf("could not proceed with deletion of %q: %s", u.Package, err)
+	}
+
+	/*
+		Deletion order:
+		1. Subscription to prevent further installs or upgrades of the operator while cleaning up.
+
+		If the CSV exists:
+		  2. Operands so the operator has a chance to handle CRs that have finalizers.
+		  Note: the correct strategy must be chosen in order to process an opertor delete with operand on-cluster.
+		  3. ClusterServiceVersion. OLM puts an ownerref on every namespaced resource to the CSV,
+		   and an owner label on every cluster scoped resource so they get gc'd on deletion.
+
+		4. OperatorGroup in the namespace if no other subscriptions are in that namespace and OperatorGroup deletion is specified
+	*/
 
 	// Subscriptions can be deleted asynchronously.
 	if err := u.deleteObjects(ctx, sub); err != nil {
 		return err
 	}
 
-	if csv != nil {
-		// Ensure CustomResourceDefinitions are deleted next, so that the operator
-		// has a chance to handle CRs that have finalizers.
-		if u.DeleteCRDs {
-			crds := getCRDs(csv)
-			if err := u.deleteObjects(ctx, crds...); err != nil {
-				return err
-			}
-		}
-
-		// OLM puts an ownerref on every namespaced resource to the CSV,
-		// and an owner label on every cluster scoped resource. When CSV is deleted
-		// kube and olm gc will remove all the referenced resources.
-		if err := u.deleteObjects(ctx, csv); err != nil {
+	// If we could not find a csv associated with the subscription, that likely
+	// means there is no CSV associated with it yet. Delete non-CSV related items only like the operatorgroup.
+	if csv == nil {
+		u.Logf("csv for package %q not found", u.Package)
+	} else {
+		if err := u.deleteCSVRelatedResources(ctx, csv, operands); err != nil {
 			return err
 		}
 	}
 
 	if u.DeleteOperatorGroups {
-		subs := v1alpha1.SubscriptionList{}
-		if err := u.config.Client.List(ctx, &subs, client.InNamespace(u.config.Namespace)); err != nil {
-			return fmt.Errorf("list subscriptions: %v", err)
-		}
-		// If there are no subscriptions left, delete the operator group(s).
-		if len(subs.Items) == 0 {
-			ogs := v1.OperatorGroupList{}
-			if err := u.config.Client.List(ctx, &ogs, client.InNamespace(u.config.Namespace)); err != nil {
-				return fmt.Errorf("list operatorgroups: %v", err)
-			}
-			for _, og := range ogs.Items {
-				og := og
-				if len(u.DeleteOperatorGroupNames) == 0 || contains(u.DeleteOperatorGroupNames, og.GetName()) {
-					if err := u.deleteObjects(ctx, &og); err != nil {
-						return err
-					}
-				}
-			}
+		if err := u.deleteOperatorGroup(ctx); err != nil {
+			return fmt.Errorf("delete operatorgroup: %v", err)
 		}
 	}
+
 	return nil
 }
 
@@ -165,28 +152,73 @@ func (u *OperatorUninstall) getSubscriptionCSV(ctx context.Context, subscription
 	return csv, name, nil
 }
 
+func (u *OperatorUninstall) deleteOperatorGroup(ctx context.Context) error {
+	subs := v1alpha1.SubscriptionList{}
+	if err := u.config.Client.List(ctx, &subs, client.InNamespace(u.config.Namespace)); err != nil {
+		return fmt.Errorf("list subscriptions: %v", err)
+	}
+
+	// If there are no subscriptions left, delete the operator group(s).
+	if len(subs.Items) == 0 {
+		ogs := v1.OperatorGroupList{}
+		if err := u.config.Client.List(ctx, &ogs, client.InNamespace(u.config.Namespace)); err != nil {
+			return fmt.Errorf("list operatorgroups: %v", err)
+		}
+		for _, og := range ogs.Items {
+			og := og
+			if len(u.DeleteOperatorGroupNames) == 0 || contains(u.DeleteOperatorGroupNames, og.GetName()) {
+				if err := u.deleteObjects(ctx, &og); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// validStrategy validates the deletion strategy against the operands on-cluster
+// TODO define and use an OperandStrategyError that the cmd can use errors.As() on to provide external callers a more generic error
+func (u *OperatorUninstall) validStrategy(operands *unstructured.UnstructuredList) error {
+	if len(operands.Items) > 0 && u.OperandStrategy.Kind == operand.Cancel {
+		return fmt.Errorf("%d operands exist and operand strategy %q is in use: "+
+			"delete operands manually or re-run uninstall with a different operand deletion strategy."+
+			"\n\nSee kubectl operator uninstall --help for more information on operand deletion strategies.", len(operands.Items), operand.Cancel)
+	}
+	return nil
+}
+
+func (u *OperatorUninstall) deleteCSVRelatedResources(ctx context.Context, csv *v1alpha1.ClusterServiceVersion, operands *unstructured.UnstructuredList) error {
+	switch u.OperandStrategy.Kind {
+	case operand.Ignore:
+		for _, op := range operands.Items {
+			u.Logf("%s %q orphaned", strings.ToLower(op.GetKind()), prettyPrint(op))
+		}
+	case operand.Delete:
+		for _, op := range operands.Items {
+			op := op
+			if err := u.deleteObjects(ctx, &op); err != nil {
+				return err
+			}
+		}
+	default:
+		return fmt.Errorf("unknown operand deletion strategy %q", u.OperandStrategy)
+	}
+
+	// OLM puts an ownerref on every namespaced resource to the CSV,
+	// and an owner label on every cluster scoped resource. When CSV is deleted
+	// kube and olm gc will remove all the referenced resources.
+	if err := u.deleteObjects(ctx, csv); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func csvNameFromSubscription(subscription *v1alpha1.Subscription) string {
 	if subscription.Status.InstalledCSV != "" {
 		return subscription.Status.InstalledCSV
 	}
 	return subscription.Status.CurrentCSV
-}
-
-// getCRDs returns the list of CRDs required by a CSV.
-func getCRDs(csv *v1alpha1.ClusterServiceVersion) (crds []client.Object) {
-	for _, resource := range csv.Status.RequirementStatus {
-		if resource.Kind == crdKind {
-			obj := &unstructured.Unstructured{}
-			obj.SetGroupVersionKind(schema.GroupVersionKind{
-				Group:   resource.Group,
-				Version: resource.Version,
-				Kind:    resource.Kind,
-			})
-			obj.SetName(resource.Name)
-			crds = append(crds, obj)
-		}
-	}
-	return
 }
 
 func contains(haystack []string, needle string) bool {
@@ -196,4 +228,12 @@ func contains(haystack []string, needle string) bool {
 		}
 	}
 	return false
+}
+
+func prettyPrint(op unstructured.Unstructured) string {
+	namespaced := op.GetNamespace() != ""
+	if namespaced {
+		return fmt.Sprint(op.GetName() + "/" + op.GetNamespace())
+	}
+	return op.GetName()
 }
--- a/internal/pkg/action/operator_uninstall_test.go
+++ b/internal/pkg/action/operator_uninstall_test.go
@@ -1,0 +1,258 @@
+package action_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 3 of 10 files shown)


internal/cmd/operator_uninstall_test.go
```diff
@@ -0,0 +1,223 @@
+package cmd_test
```
joelanford, line 1: I don't see any actual usage of the `internal/cmd` package in this test. It seems more appropriate to put this file (and ` internal/cmd/cmd_suite_test.go`) in `internal/action` instead.

internal/pkg/action/operator_uninstall.go
```diff
@@ -5,33 +5,29 @@ import (
...
-
-	Package                  string
-	DeleteAll                bool
-	DeleteCRDs               bool
-	DeleteOperatorGroups     bool
-	DeleteOperatorGroupNames []string
+	Package         string
+	config          *action.Configuration
+	operandLister   *action.OperatorListOperands
+	OperandStrategy operand.DeletionStrategy
```
joelanford, line 21: Nit: keep the unexported fields separate from the exported fields, and try to keep both sets of fields in an API-user-friendly ordering (e.g. in decreasing order of how often we expect the field to be set or used)

internal/pkg/action/operator_uninstall.go
```diff
@@ -5,33 +5,29 @@ import (
...
 
 	Logf func(string, ...interface{})
 }
 
 func NewOperatorUninstall(cfg *action.Configuration) *OperatorUninstall {
 	return &OperatorUninstall{
-		config: cfg,
-		Logf:   func(string, ...interface{}) {},
+		config:        cfg,
+		operandLister: action.NewOperatorListOperands(cfg),
```
joelanford, line 29: Any reason we need a separate struct field for `operandLister` right now vs. instantiating it during `Run`? I try to keep structs as minimal as possible and only add fields when it's clear there is a need.

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,10 +65,20 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
 		return fmt.Errorf("get subscription CSV %q: %v", csvName, err)
 	}
 
+	if csv == nil {
+		return fmt.Errorf("get subscription CSV %q: %v", csvName, err)
+	}
```
joelanford, line 70: I'm not sure I agree with this being an error condition. See https://github.com/operator-framework/kubectl-operator/blob/9e5901c77b37c9a709cedc080733d3ca7bb50b64/internal/pkg/action/operator_uninstall.go#L148-L153

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+		return fmt.Errorf("unknown operand deletion strategy provided: %s", u.OperandStrategy.String())
+	}
+
+	// OLM puts an ownerref on every namespaced resource to the CSV,
+	// and an owner label on every cluster scoped resource. When CSV is deleted
+	// kube and olm gc will remove all the referenced resources.
+	if err := u.deleteObjects(ctx, csv); err != nil {
+		return err
 	}
+
```
joelanford, line 122: See above comment about a `nil` CSV being a non-error case.

The scenario is that a user runs `kubectl operator install`, which creates a subscription that doesn't work for whatever reason. They should see that there was an error and then be able to run `kubectl operator uninstall` to basically undo the effect of the previously run `install` command.

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
-			if err := u.config.Client.List(ctx, &ogs, client.InNamespace(u.config.Namespace)); err != nil {
-				return fmt.Errorf("list operatorgroups: %v", err)
-			}
-			for _, og := range ogs.Items {
-				og := og
-				if len(u.DeleteOperatorGroupNames) == 0 || contains(u.DeleteOperatorGroupNames, og.GetName()) {
-					if err := u.deleteObjects(ctx, &og); err != nil {
-						return err
-					}
-				}
```
joelanford, line 124: I'm not convinced we should get rid of this. It seems pretty useful, and I think is somewhat orthogonal to the rest of this PR. Could we keep this for this PR and have a more focused issue/PR for operator group deletion in a follow-up?

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+	u.Logf("operand deletion strategy: %s", u.OperandStrategy.String())
 
-	if u.DeleteOperatorGroups {
-		subs := v1alpha1.SubscriptionList{}
-		if err := u.config.Client.List(ctx, &subs, client.InNamespace(u.config.Namespace)); err != nil {
-			return fmt.Errorf("list subscriptions: %v", err)
+	switch u.OperandStrategy {
+	case operand.Cancel:
+		if len(operands.Items) > 0 {
+			return fmt.Errorf("deleting an operator with operands on-cluster requires a operand deletion strategy to be set")
```
joelanford, line 100:
```suggestion
			return fmt.Errorf("cannot uninstall operator when operands exist and operand strategy %q is in use; delete operands or re-run with a different operand strategy", operand.Cancel)
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
-		// kube and olm gc will remove all the referenced resources.
-		if err := u.deleteObjects(ctx, csv); err != nil {
-			return err
-		}
+	operands, err := u.operandLister.Run(ctx, u.Package)
+	if err != nil {
+		return fmt.Errorf("listing operands associated with operator %s: %s", u.Package, err)
 	}
+	u.Logf("found %d operands associated with operator %s", len(operands.Items), u.Package)
+	u.Logf("operand deletion strategy: %s", u.OperandStrategy.String())
```
joelanford, line 95: I'd suggest removing these logs, and sticking to `<verb>ing <gvk> <name>/<namespace>` style logs to maintain consistency.

Eg.
- `deleting <gvk> <name>/<namespace>`
- `orphaning <gvk> <name>/<namespace>`

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
-			}
-			for _, og := range ogs.Items {
-				og := og
-				if len(u.DeleteOperatorGroupNames) == 0 || contains(u.DeleteOperatorGroupNames, og.GetName()) {
-					if err := u.deleteObjects(ctx, &og); err != nil {
-						return err
-					}
-				}
+	case operand.Ignore:
+		u.Logf("ignoring %d operands on-cluster managed by operator %s", len(operands.Items), u.Package)
```
joelanford, line 103: Iterate the items and list each one out with an `ophaned` verb?

internal/pkg/action/operator_uninstall.go
```diff
@@ -86,45 +87,39 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+	case operand.Delete:
+		for _, op := range operands.Items {
+			// TODO parallelize?
+			op := op
+			if err := u.deleteObjects(ctx, &op); err != nil {
+				return err
 			}
 		}
+	default:
+		return fmt.Errorf("unknown operand deletion strategy provided: %s", u.OperandStrategy.String())
```
joelanford, line 113: Nit: keep it short and sweet
Suggestion: use `%q` when printing strings that may be empty or are user input. `%s` with empty strings is a pretty crummy UX
```suggestion
		return fmt.Errorf("unknown operand deletion strategy %q", u.OperandStrategy)
```

internal/pkg/operand/strategy.go
```diff
@@ -0,0 +1,24 @@
...
+	// Ignore will ignore the operands when deleting the operator, in effect orphaning them.
+	Ignore DeletionStrategy = "ignore"
+	// Delete will delete the operands associated with the operator before deleting the operator, allowing finalizers to run.
+	Delete DeletionStrategy = "delete"
+)
+
+func (d DeletionStrategy) String() string {
+	return string(d)
+}
+
```
joelanford, line 18: Since `DeletionStrategy` is already a string, this function seems unnecessary. If we ever need to have a custom `String()` function, I think we can add it later.

internal/pkg/operand/strategy.go
```diff
@@ -0,0 +1,24 @@
...
+func (d DeletionStrategy) Valid() bool {
+	if d != Cancel && d != Ignore && d != Delete {
+		return false
+	}
+	return true
```
joelanford, line 23: Nit: a switch makes this a little nicer:
```suggestion
	switch d {
	case Cancel, Ignore, Delete:
		return true
	}
	return false
```

internal/pkg/operand/strategy.go
```diff
@@ -0,0 +1,24 @@
+package operand
+
+// DeletionStrategy describes how to handle operands on-cluster when deleting the associated operator.
+type DeletionStrategy string
```
joelanford, line 4: It would be nice to make this type implement the `pflag.Value` interface. Then we could plug it in directly and we could perform validation during flag parsing.

internal/cmd/operator_uninstall_test.go
```diff
@@ -0,0 +1,223 @@
...
+		cfg.Client = cl
+		cfg.Namespace = "etcd-namespace"
+	})
+
+	It("should fail due to missing subscription", func() {
+		uninstaller := internalaction.NewOperatorUninstall(&cfg)
+		// switch to package without a subscription for it
+		uninstaller.Package = "redis"
+		err := uninstaller.Run(context.TODO())
+		Expect(err.Error()).To(ContainSubstring("package " + `"` + "redis" + `"` + " not found"))
```
joelanford, line 153:
```suggestion
		Expect(err).To(MatchError(&internalaction.ErrPackageNotFound{PackageName: "redis"}))

```

internal/pkg/action/operator_uninstall.go
joelanford: It might be worthwhile to move the sorting we currently do [here](https://github.com/operator-framework/kubectl-operator/blob/9e5901c77b37c9a709cedc080733d3ca7bb50b64/internal/cmd/operator_list_operands.go#L67-L78) into the end of the  list operands `Run` function.

internal/pkg/action/operator_uninstall.go
exdx: Makes sense to me 👍

internal/pkg/action/operator_uninstall.go
exdx: Fixed to return a log and a nil error instead

internal/pkg/action/operator_uninstall.go
exdx: This makes sense but would break the `OperandLister` when it can't find the CSV associated with the operator object on-cluster. We can move the subscription deletion back up higher in the code to still be able to delete a subscription even if there is no CSV.

internal/pkg/action/operator_uninstall.go
exdx: We could break that out into a separate PR, sure. 

I feel pretty strongly that we should only delete the minimal subset of items for this command for the initial downstream build, mainly because operatorgroups are not well understood by most OLM users. If a user does understand operatorgroups, they should feel comfortable deleting them manually, even though it can have implications for subsequent installations.

The scenario I'd like to avoid is having a user inadvertently delete the global operatorgroup in the `openshift-operators` namespace with the `uninstall` command and then have all subsequent operator installations fail.

internal/pkg/action/operator_uninstall.go
exdx: sounds good

internal/pkg/operand/strategy.go
exdx: That sounds great

joelanford:
> how do we reconcile the fact that kubectl operator uninstall leaves the Operator object associated with the CSV on-cluster if the operator installs CRDs

In my opinion, I think its okay to leave the `Operator` object in place on-cluster as a breadcrumb for finding CRDs. If I remember correctly, `kubectl operator uninstall` does not currently delete the `Operator` object, so this PR would not revert that aspect of the existing functionality. 

In follow-on issues we could cover CRD and Operator deletion strategies as separate concerns.

internal/pkg/action/operator_uninstall.go
joelanford: > mainly because operatorgroups are not well understood by most OLM users

This cuts both ways. With `install` creating an operatorgroup (if necessary), and `uninstall` deleting the operatorgroup (when possible), the commands have symmetry and the existing UX means users _can_ be completely unaware of what they are and how they work and still interact with OLM.

> The scenario I'd like to avoid is having a user inadvertently delete the global operatorgroup in the openshift-operators namespace with the uninstall command and then have all subsequent operator installations fail.

This is not a scenario I had considered, but I agree it should be. Is it currently possible for a user with the necessary RBAC permissions to delete the global operatorgroup in the `openshift-operators` namespace permanently? Or will some controller re-create it when it finds that it is missing?

exdx:
> > how do we reconcile the fact that kubectl operator uninstall leaves the Operator object associated with the CSV on-cluster if the operator installs CRDs
> 
> In my opinion, I think its okay to leave the `Operator` object in place on-cluster as a breadcrumb for finding CRDs. If I remember correctly, `kubectl operator uninstall` does not currently delete the `Operator` object, so this PR would not revert that aspect of the existing functionality.
> 
> In follow-on issues we could cover CRD and Operator deletion strategies as separate concerns.

SGTM

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,41 +66,68 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+	if err := u.deleteObjects(ctx, sub); err != nil {
+		return err
+	}
+
+	// If we could not find a csv associated with the subscription, that likely
+	// means there is no CSV associated with it yet. Return a nil error.
+	if csv == nil {
+		u.Logf("csv for package %q not found", u.Package)
+		return nil
+	}
```
joelanford, line 79: Returning early here means we don't get a chance to see if we can delete operator groups or other bundle artifacts that may have been deployed separately from the CSV (e.g. eventually CRDs, hopefully). We will likely change operator group handling in a follow-on PR, but in general, I think we should gracefully handle missing CSVs and just skip the steps that require them.

As an example, I would expect the following output if an install failed to deploy an operator because of a non-existent version. In this case a CSV never gets created, but uninstall figures out what's present and deletes what exists:
```console
$ kubectl operator install etcd -v 1.2.3 -w etcd --create-operator-group
operatorgroup "etcd" created
subscription "etcd" created
failed to install operator: waiting for install plan to exist: timed out waiting for the condition

$ kubectl get operatorgroup,subscription
NAME                                      AGE
operatorgroup.operators.coreos.com/etcd   86s

NAME                                     PACKAGE   SOURCE                  CHANNEL
subscription.operators.coreos.com/etcd   etcd      operatorhubio-catalog

$ kubectl operator uninstall etcd --delete-operator-groups
subscription "etcd" deleted
operatorgroup "etcd" deleted
operator "etcd" uninstalled
```

joelanford: For the list-operands problem, we should talk more about what should happen when a CSV is not present. It is possible to have operands without the CSV. But it is currently not possible to know what operands are associated with an operator without a CSV.

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,41 +66,68 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+
 	// Deletion order:
 	//
 	// 1. Subscription to prevent further installs or upgrades of the operator while cleaning up.
-	// 2. CustomResourceDefinitions so the operator has a chance to handle CRs that have finalizers.
+	// 2. Operands so the operator has a chance to handle CRs that have finalizers.
+	// Note: the correct strategy must be chosen in order to process an opertor delete with operand on-cluster.
 	// 3. ClusterServiceVersion. OLM puts an ownerref on every namespaced resource to the CSV,
 	//    and an owner label on every cluster scoped resource so they get gc'd on deletion.
```
joelanford, line 87: Nit: move these comment lines above the subscription deletion.

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,41 +66,68 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
 
-	// Subscriptions can be deleted asynchronously.
-	if err := u.deleteObjects(ctx, sub); err != nil {
-		return err
+	// create lister and list operands
+	lister := action.NewOperatorListOperands(u.config)
+	operands, err := lister.Run(ctx, u.Package)
+	if err != nil {
+		return fmt.Errorf("listing operands associated with operator %s: %s", u.Package, err)
```
joelanford, line 93: Nit: 
```suggestion
		return fmt.Errorf("listing operands associated with operator %q: %v", u.Package, err)
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,41 +66,68 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
-			if err := u.deleteObjects(ctx, crds...); err != nil {
+	switch u.OperandStrategy.Kind {
+	case operand.Cancel:
+		if len(operands.Items) > 0 {
+			return fmt.Errorf("%d operands exist and operand strategy %q is in use; "+
+				"delete operands or re-run with a different operand strategy", len(operands.Items), operand.Cancel)
+		}
+	case operand.Ignore:
+		for _, op := range operands.Items {
+			u.Logf("%s %q orphaned", strings.ToLower(op.GetKind()), op.GetName()+"/"+op.GetNamespace())
```
joelanford, line 104: Some operands will be cluster-scoped and therefore not have namespaces. We should handle that case separately instead of printing something like `clusterissuer "my-cluster-issuer/" orphaned`.

Perhaps there a helper type in a k8s library that can handle this for us?

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,41 +66,68 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+		}
+	case operand.Delete:
+		for _, op := range operands.Items {
+			op := op
+			if err := u.deleteObjects(ctx, &op); err != nil {
 				return err
 			}
+			u.Logf("%s %q deleted", strings.ToLower(op.GetKind()), op.GetName()+"/"+op.GetNamespace())
```
joelanford, line 112: I _think_ `deleteObjects()` already logs this line. If so no need to repeat it.

internal/pkg/action/operator_uninstall.go
exdx: Ok, thanks for that example, I think I'm following now. I will look to add this in

internal/pkg/action/operator_uninstall.go
```diff
@@ -74,57 +67,77 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
 
-	if csv != nil {
-		// Ensure CustomResourceDefinitions are deleted next, so that the operator
-		// has a chance to handle CRs that have finalizers.
-		if u.DeleteCRDs {
-			crds := getCRDs(csv)
-			if err := u.deleteObjects(ctx, crds...); err != nil {
-				return err
+	// If we could not find a csv associated with the subscription, that likely
+	// means there is no CSV associated with it yet. Delete non-CSV related items and return early.
```
joelanford, line 89: It would be better to avoid duplication of the operator group deletion code.

Could we put the CSV-related stuff into a separate function and then have something like:
```go
if csv == nil {
    u.Logf("csv for package %q not found", u.Package)
} else {
    if err := u.deleteCSVRelatedResources(ctx); err != nil {
        return err
    }
}
// do operatorgroup deletion stuff down here
```

internal/pkg/action/operator_uninstall.go
```diff
@@ -165,30 +178,37 @@ func (u *OperatorUninstall) getSubscriptionCSV(ctx context.Context, subscription
 	return csv, name, nil
 }
 
+func (u *OperatorUninstall) deleteOperatorGroup(ctx context.Context) error {
```
joelanford, line 181: :+1:

joelanford (commented):
Looking great. Pretty much down to nits around consistent log style, and one suggestion to reduce duplication.

internal/pkg/action/operator_uninstall_test.go
```diff
@@ -0,0 +1,248 @@
...
+		Expect(cfg.Client.Update(context.TODO(), sub)).To(Succeed())
+
+		uninstaller := internalaction.NewOperatorUninstall(&cfg)
+		uninstaller.Package = etcd
+		err := uninstaller.Run(context.TODO())
+		Expect(err).To(BeNil())
+
+		subKey := types.NamespacedName{Name: "etcd-sub", Namespace: "etcd-namespace"}
+		s := &v1alpha1.Subscription{}
+		Expect(cfg.Client.Get(context.TODO(), subKey, s)).To(WithTransform(apierrors.IsNotFound, BeTrue()))
```
benluddy, line 170: Can you bump the gomega dependency to get https://github.com/onsi/gomega/pull/423? Otherwise this will panic instead of failing when Get returns nil.

timflannagan (commented):
Just a quick drive-by while checking if tide had been rolled out yet.

internal/cmd/operator_uninstall.go
```diff
@@ -15,8 +15,26 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
...
+		Long: `Uninstall removes the subscription, operator and optionally operands managed by the operator as well as the relevant operatorgroup.
+
+Warning: this command permanently deletes objects from the cluster. Running uninstall concurrently with other operations
+could result in undefined behavior. 
+
+The uninstall command first finds and deletes the subscription associated with the operator. It then 
+lists all operands found throughout the cluster for the operator
+specified if one is found. Since the scope of an operator is restricted by
+its operator group, this search will include namespace-scoped operands from the
+operator group's target namespaces and all cluster-scoped operands.
```
joelanford, line 28: Might need to make a slight tweak to this description since we check the operands and operand strategy before deleting anything now.

internal/pkg/action/operator_uninstall.go
```diff
@@ -165,28 +152,72 @@ func (u *OperatorUninstall) getSubscriptionCSV(ctx context.Context, subscription
...
-				Kind:    resource.Kind,
-			})
-			obj.SetName(resource.Name)
-			crds = append(crds, obj)
+// validStrategy validates the deletion strategy against the operands on-cluster
+func (u *OperatorUninstall) validStrategy(operands *unstructured.UnstructuredList) (bool, error) {
+	if len(operands.Items) > 0 && u.OperandStrategy.Kind == operand.Cancel {
+		return false, fmt.Errorf("%d operands exist and operand strategy %q is in use: "+
+			"delete operands manually or re-run uninstall with a different operand deletion strategy."+
+			"\n\nSee kubectl operator uninstall --help for more information on operand deletion strategies.", len(operands.Items), operand.Cancel)
```
joelanford, line 184: This is fine for now, but I think we should follow up and rework the way this error message works. Theoretically, once we export the `OperatorUninstall` struct in a non-internal tree, this function could be imported and used by other tools (e.g. operator-sdk's `cleanup` subcommand), so it would be better not reference the CLI in this package.

I think we could rework this by defining an error struct e.g. `OperandStrategyError` that the cmd layer can use `errors.As` on, and we could then put the CLI-related log messages in the cmd layer instead of the `action` package.

Could you add a TODO comment as a reminder to ourselves?

internal/pkg/action/operator_uninstall_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	It("should fail due to missing subscription", func() {
+		uninstaller := internalaction.NewOperatorUninstall(&cfg)
+		// switch to package without a subscription for it
+		uninstaller.Package = "redis"
+		err := uninstaller.Run(context.TODO())
+		Expect(err).To(MatchError(&internalaction.ErrPackageNotFound{PackageName: "redis"}))
+	})
+
+	It("should fail due to missing csv", func() {
```
joelanford, line 157: It doens't fail in this case anymore right? Should we change the name of the test?

internal/pkg/action/operator_uninstall_test.go
```diff
@@ -0,0 +1,248 @@
...
+		err := uninstaller.Run(context.TODO())
+		Expect(err).To(BeNil())
+
+		subKey := types.NamespacedName{Name: "etcd-sub", Namespace: "etcd-namespace"}
+		s := &v1alpha1.Subscription{}
+		Expect(cfg.Client.Get(context.TODO(), subKey, s)).To(WithTransform(apierrors.IsNotFound, BeTrue()))
+
+		csvKey := types.NamespacedName{Name: "etcdoperator.v0.9.4-clusterwide", Namespace: "etcd-namespace"}
+		csv := &v1alpha1.ClusterServiceVersion{}
+		Expect(cfg.Client.Get(context.TODO(), csvKey, csv)).To(WithTransform(apierrors.IsNotFound, BeTrue()))
```
joelanford, line 202: Also check that operands still exist in this case?

internal/pkg/action/operator_uninstall.go
exdx: Makes total sense

exdx:
squashed commits

internal/pkg/action/operator_uninstall.go
```diff
@@ -69,62 +62,56 @@ func (u *OperatorUninstall) Run(ctx context.Context) error {
...
+	// find operands related to the operator on cluster
+	lister := action.NewOperatorListOperands(u.config)
+	operands, err := lister.Run(ctx, u.Package)
+	if err != nil {
+		return fmt.Errorf("list operands for operator %q: %v", u.Package, err)
+	}
+	// validate the provided deletion strategy before proceeding to deletion
+	if ok, err := u.validStrategy(operands); !ok {
+		return fmt.Errorf("could not proceed with deletion of %q: %s", u.Package, err)
+	}
```
joelanford, line 79: Nit: It isn't possible with the current implementation, but if `validStrategy()` returns `false, nil` we'll get a strange looking error message.

Perhaps we can just make this function return just an `error`?
```suggestion
	if err := u.validateStrategy(operands); err != nil {
		return fmt.Errorf("could not proceed with deletion of %q: %s", u.Package, err)
	}
```

internal/cmd/operator_uninstall.go
```diff
@@ -15,8 +15,26 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
...
+its operator group, this search will include namespace-scoped operands from the
+operator group's target namespaces and all cluster-scoped operands. 
+
+The operand-deletion strategy is then considered if any operands are found on-cluster. One of cancel|ignore|delete. 
+By default, the strategy is "cancel", which means that if any operands are found when deleting the operator abort the uninstall without deleting anything. 
+The "ignore" strategy keeps the operands on cluster and deletes the subscription and the operator.
+The "delete" strategy deletes the subscription, operands, and after they have finished finalizing, the operator itself.
+
+Setting --delete-operator-groups to true will delete the operatorgroup in the provided namespace if no other active subscriptions 
+are currently in that namespace, after removing the operator. The subscription and operatorgroup will be removed even if the operator is not found.`,
```
joelanford, line 36: Should we have a consistent line break to make this format well on most consoles (perhaps cap it at 80 or 100 characters?)

This makes me wonder if there's a library that help us do this consistently across all commands.

joelanford (commented):
My final couple of comments. After this I'll be ready to merge.

internal/cmd/operator_uninstall.go
exdx: Per https://github.com/spf13/cobra/issues/484 looks like we should just use strings instead of string literals?

joelanford: I think the paragraphs and line-splitting with string literals like you have in the current long description are fine. My only complaint is that there does not seem to be a consistent line width.

I would have expected something like:
```suggestion
		Long: `Uninstall removes the subscription, operator and optionally operands managed by
the operator as well as the relevant operatorgroup.

Warning: this command permanently deletes objects from the cluster. Running
uninstall concurrently with other operations could result in undefined
behavior. 

The uninstall command first checks to find the subscription associated with the
operator. It then lists all operands found throughout the cluster for the
operator specified if one is found. Since the scope of an operator is restricted
by its operator group, this search will include namespace-scoped operands from
the operator group's target namespaces and all cluster-scoped operands. 

The operand-deletion strategy is then considered if any operands are found
on-cluster. One of cancel|ignore|delete. 
  - By default, the strategy is "cancel", which means that if any operands are
    found when deleting the operator abort the uninstall without deleting
    anything. 
  - The "ignore" strategy keeps the operands on cluster and deletes the
    subscription and the operator.
  - The "delete" strategy deletes the subscription, operands, and after they have
    finished finalizing, the operator itself.

Setting --delete-operator-groups to true will delete the operatorgroup in the
provided namespace if no other active subscriptions are currently in that
namespace, after removing the operator. The subscription and operatorgroup will
be removed even if the operator is not found.`,
```

PR #52, opened 2021-08-03
operator install: honor operatorframework.io/suggested-namespace

### LINKED ISSUE #50: Feature req: Install operator into namespace
I have been working on an operator which is listed in the Operator Hub catalog, and it has some resources (clusterrolebinding) which are expected to be created for service accounts in a certain namespace. The operator uses `operatorframework.io/suggested-namespace` to cajole the OpenShift console to install it into that namespace.

When using OLM without OpenShift, I have found that suggested-namespace is not apparently honored. The result of this is some ServiceAccounts are created in the `operators` namespace, and the ClusterRoleBindings are bound to service account in the suggested namespace that do not exist. (It was easy enough to update the ClusterRoleBindings by hand, but this is easier done than said, in other words it will be a bad mark on our documentation if we have to explain this...)

While in the long term the plan is to move away from service accounts and impersonate users instead, I'm not sure how to handle this in the short term. I understood that the OpenShift Console can be used outside of OpenShift, and that can be used to install operators and create namespaces, but that Operator Lifecycle Manager will not create namespaces by itself.

Would this be a reasonable feature addition to the kubectl-operator plugin? Is there somewhere else it should go besides OpenShift Console?

I haven't been through every page of the docs, but is there already a nice way to install the operator into a user-specified namespace at install time? (Would it help if I can create the namespace some other way?) I'm looking for the best way to resolve this, and someone suggested this kubectl plugin project might be a good place for this function to land.

Skimming through the options, I found there is a `-n` flag for namespace, but it does not appear to have the desired effect.

```
kubectl operator install -n flux-system flux --create-operator-group
```

(Flux still lands in the `operators` namespace.)

joelanford:
**Description:**
Updates `kubectl operator install` to honor CSVs that provide an `operatorframework.io/suggested-namespace` annotation.

**Motivation:**
Closes #50

/hold
I can't find any upstream documentation for `operatorframework.io/suggested-namespace`, so I'm somewhat concerned that we're adding support for a downstream feature. There's nothing actually OpenShift-specific with this annotation other than that OpenShift Console is the only _current_ client that supports it. I think this would be as simple as a doc change to upstream OLM docs that specifies this annotation and describes what clients should do when it's present. 

Downstream enhancement describes this annotation here: https://github.com/openshift/enhancements/blob/master/enhancements/olm/olm-managed-operator-metrics.md#fulfilling-namespace-and-rbac-requirements

### DIFF
--- a/internal/pkg/action/operator_install.go
+++ b/internal/pkg/action/operator_install.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
@@ -10,6 +11,7 @@ import (
 	v1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -23,13 +25,14 @@ import (
 type OperatorInstall struct {
 	config *action.Configuration
 
-	Package             string
-	Channel             string
-	Version             string
-	Approval            subscription.ApprovalValue
-	WatchNamespaces     []string
-	CleanupTimeout      time.Duration
-	CreateOperatorGroup bool
+	Package                  string
+	Channel                  string
+	Version                  string
+	Approval                 subscription.ApprovalValue
+	WatchNamespaces          []string
+	CleanupTimeout           time.Duration
+	CreateOperatorGroup      bool
+	PermitAlternateNamespace bool
 
 	Logf func(string, ...interface{})
 }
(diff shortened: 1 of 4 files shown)


internal/pkg/action/operator_install.go
```diff
@@ -110,6 +128,18 @@ func (i *OperatorInstall) getPackageManifest(ctx context.Context) (*operator.Pac
 	return &operator.PackageManifest{PackageManifest: *pm}, nil
 }
 
+func (i *OperatorInstall) ensureNamespace(ctx context.Context, pc *operator.PackageChannel) error {
+	suggestedNamespace := pc.CurrentCSVDesc.Annotations["operatorframework.io/suggested-namespace"]
```
estroz, line 132: This could panic I believe.

joelanford: It can panic if `pc` is `nil`, but that can't happen because `pm.GetChannel()` (called one function up in the stack) doesn't return `nil, nil`

Other than that, it can't panic:
- `pc.CurrentCSVDesc` is a concrete struct
- `pc.CurrentCSVDesc.Annotations` is a map[string]string

If you do a lookup of a key in a nil map, the returned value is the default value of the value type.

estroz: >If you do a lookup of a key in a nil map, the returned value is the default value of the value type.

TIL! This is kinda crazy to me because a [map is really a pointer to an hmap](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it), so you're dereferencing and indexing a nil pointer.

perdasilva:
closing due to inactivity - please re-open if needed ^^

PR #55, opened 2021-08-27
Uninstall improvements

joelanford:
This PR adds flags and functionality for deleting more operator-related objects
- `--delete-operator` - delete the operator object and everything it references, implies `--operand-strategy=delete`.
- `--delete-all` - basically `--delete-operator --delete-operator-groups`

It also simplifies the `operand.DeletionStrategy` type to be a string instead of a struct.

Lastly, it adds a specific error for `CancelStrategyError` to decouple the uninstall CLI from the uninstall action. This enables moving CLI-specific logging to the CLI layer.

---

We can probably debate the UX for the re-added deletion flags, but ever since we've removed `--delete-all` on the `main` branch, I've found myself rebuilding from the v0.3.0 tag so I could keep using it.

The frequency that I personally do `kubectl operator install`, run a few tests, and then do `kubectl operator uninstall -X` tells me that it is a very useful option to have, and I think that merits adding it back in, at least until we come up with an alternative that maintains a simple install, test, wipe clean type of workflow.

### DIFF
--- a/internal/cmd/operator_uninstall.go
+++ b/internal/cmd/operator_uninstall.go
@@ -1,11 +1,14 @@
 package cmd
 
 import (
+	"errors"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
 	internalaction "github.com/operator-framework/kubectl-operator/internal/pkg/action"
+	"github.com/operator-framework/kubectl-operator/internal/pkg/operand"
 	"github.com/operator-framework/kubectl-operator/pkg/action"
 )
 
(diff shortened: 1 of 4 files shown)


internal/cmd/operator_uninstall.go
```diff
@@ -16,42 +19,70 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
...
+      and then deleting the operator object itself. This implies the flag
+      '--operand-strategy=delete' because it is impossible to delete CRDs
+      without also deleting instances of those CRDs.
+
+  -X, --delete-all
+
+      This is a convenience flag that is effectively equivalent to the flags
+      '--delete-operator=true --delete-operator-groups=true'. This is
+      most useful to completely undo the effect of a previous run of
+      'kubectl operator install <operator> --create-operator-group'
```
joelanford, line 64: This isn't _quite_ true in all cases.

If an operator (A) depends on another operator (B), and I install A on a clean cluster, OLM will also ensure B gets installed.

The `uninstall` command will not automatically uninstall B.

To do that, the plugin would need to (at uninstall time) know that:
a) B is a dependency of A
b) B was installed automatically by OLM (i.e. not manually by a user)
c) no other installed operators depend on B

We could do something similar to `apt-get autoremove` which is a separate command that uninstalls unused dependencies.

But for now, I'll update this description to explicitly state that dependent operators must be uninstalled separately.

internal/cmd/operator_uninstall.go
```diff
@@ -16,42 +19,73 @@ func newOperatorUninstallCmd(cfg *action.Configuration) *cobra.Command {
...
+uninstall concurrently with other operations could result in undefined behavior.
+
+The uninstall command first checks to find the subscription associated with the
+operator. It then lists all operands found throughout the cluster for the
+operator specified if one is found. Since the scope of an operator is restricted
+by its operator group, this search will include namespace-scoped operands from
+the operator group's target namespaces and all cluster-scoped operands.
+
+The operand-deletion strategy is then considered if any operands are found
+on-cluster. One of cancel|ignore|delete. By default, the strategy is "cancel",
```
estroz, line 35: WDYT about "abort" over "cancel"? For some reason I find cancel less easy to associate with what's happening than abort.

joelanford: I _much_ prefer abort. However, there _might_ already be external use of `cancel` that would make such a change more difficult. I'll look into it.

joelanford: The only stakeholder I know of for this enum name is the OpenShift Console, but it appears as if that project is using just the ListOperands action, and _not_ the kubectl operator uninstall CLI or action.

Therefore, I think this change is fine.

estroz (approved):

PR #73, opened 2022-04-28
Allow installing into global-operator group

kouk:
The possibleInstallModes method doesn't deal with the special case of a
target namespace being the empty string, so validation fails when
installing to the global operator group. E.g.

    ❯ kubectl operator install cert-manager --channel stable --approval Automatic -n operators
    failed to install operator: operator "cert-manager" not installable: install modes supported by operator ("AllNamespaces") not compatible with install modes supported by existing operator group ("SingleNamespace")

Add a special method to determine the operator group's supported
methods. Ref: https://olm.operatorframework.io/docs/concepts/crds/operatorgroup/#installmodes-and-supported-operatorgroups

### DIFF
--- a/internal/pkg/action/operator_install.go
+++ b/internal/pkg/action/operator_install.go
@@ -81,6 +81,21 @@ func (i *OperatorInstall) Run(ctx context.Context) (*v1alpha1.ClusterServiceVers
 	return csv, nil
 }
 
+func (i *OperatorInstall) operatorGroupSupportedInstallModes(og v1.OperatorGroup) sets.String {
+	ogNamespace := og.ObjectMeta.Namespace
+	ogTargetNamespaces := og.Status.Namespaces
+	if len(ogTargetNamespaces) == 1 {
+		if ogTargetNamespaces[0] == "" {
+			return sets.NewString(string(v1alpha1.InstallModeTypeAllNamespaces))
+		}
+		if ogTargetNamespaces[0] == ogNamespace {
+			return sets.NewString(string(v1alpha1.InstallModeTypeOwnNamespace))
+		}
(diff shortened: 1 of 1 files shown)


internal/pkg/action/operator_install.go
```diff
@@ -81,6 +81,21 @@ func (i *OperatorInstall) Run(ctx context.Context) (*v1alpha1.ClusterServiceVers
 	return csv, nil
 }
 
+func (i *OperatorInstall) operatorGroupSupportedInstallModes(og v1.OperatorGroup) sets.String {
```
joelanford, line 84: Rather than adding an entirely new function, it would be better to update the `possibleInstallModes` function to handle the case where `os.Status.Namespaces[0] == ""`

Something like:
```go
        case 1:
                switch watchNamespaces[0] {
                case "":
                        return sets.NewString(string(v1alpha1.InstallModeTypeAllNamespaces))
                case i.config.Namespace:
                        return sets.NewString(string(v1alpha1.InstallModeTypeOwnNamespace))
                default:
                        return sets.NewString(string(v1alpha1.InstallModeTypeSingleNamespace))
                }
```

joelanford:
Hi @kouk, I think the issue you were experiencing may have been solved by https://github.com/operator-framework/kubectl-operator/pull/80

On a build from `main`, I retried your command and got this:
```
$ kubectl operator install cert-manager --channel stable --approval Automatic -n operators
subscription "cert-manager" created
operator "cert-manager" installed; installed csv is "cert-manager.v1.10.2"
```

I'm going to close this out since it looks like its fixed, but please open another issue or PR if you see something else missing!

kouk:
thanks @joelanford I can't check it right now but seems legit.

PR #74, opened 2022-05-13
assets/demo: Use the eval built-in command for typed commands

timflannagan:
Use the `eval` built-in bash command to handle the typed command strings. This change should allow more complex commands that use pipe or redirect output.

Signed-off-by: timflannagan <timflannagan@gmail.com>

### DIFF
--- a/assets/demo/demo.sh
+++ b/assets/demo/demo.sh
@@ -26,7 +26,7 @@ type() {
 	done
 	echo ""
 	sleep 0.25
-	$1
+	eval $1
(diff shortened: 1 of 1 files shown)


timflannagan:
We don't have to merge this but it was just something I noticed locally in case others found it useful.

timflannagan:
The go-apidiff check is failing - opened https://github.com/operator-framework/kubectl-operator/pull/74 which should hopefully fix that issue.

PR #78, opened 2022-11-22
Return only found PackageManifests

cdjohnson:
Fixing:
```
kubectl-operator describe myoperator
package "" does not have a default channel
```


joelanford:
/lgtm
/approve

I noticed that passing a non-existent package name results in a "index out-of-range" panic (not caused by this PR). So let's get this merged, and then I'll make a follow-up to fix the panic.

PR #84, opened 2023-01-18
Add experimental OLMv1 support behind envvar toggle

joelanford:
Closes #83

PR #85, opened 2023-01-25
OLMv1 Operator install/uninstall

### LINKED ISSUE #83: Add experimental OLMv1 support in kubectl-operator for install/uninstall
There has been some recent progress with OLMv1 and the new Operator API that would be useful to explore here. We can use the kubectl-operator CLI to help drive demos and provide feedback to core OLMv1 maintainers about the UX of the new API.

Acceptance criteria
* Experimental OLMv1 support is hidden and disabled by default
* Experimental OLMv1 support is enabled via an environment variable
* When OLMv1 support is enabled, OLMv0 commands are disabled
* `operator install <packageName>` creates an `Operator` object with `metadata.name` and `spec.packageName` set to the provided package name
* `operator uninstall <packageName>` deletes the named `Operator` object

tmshort:
Fixes #83

Implements Jira OLM-2863 and OLM-2869

Signed-off-by: Todd Short <tshort@redhat.com>


tmshort:
Based on #84

internal/pkg/action/operator_install.go
```diff
@@ -7,11 +7,11 @@ import (
 	"strings"
 	"time"
 
+	sets "github.com/noahschumacher/go-collections/collections"
 	v1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/sets"
```
joelanford, line 14: Curious why we're switching from this package and introducing a new dependency?

joelanford (commented):
Also interested in a convo weighing pros/cons of separate subcommand vs different CLI based on envvar.

Part of my envvar motivation was that we would eventually flip the default to the OLMv1 CLI, in which case no one using it already would need to change their invocations.

Con of that, I suppose, is discoverability, which I was banking on relying on docs for.

internal/pkg/action/operator_install.go
tmshort: The go linter didn't like `sets.String`.

tmshort:
> Also interested in a convo weighing pros/cons of separate subcommand vs different CLI based on envvar.
> 
> Part of my envvar motivation was that we would eventually flip the default to the OLMv1 CLI, in which case no one using it already would need to change their invocations.
> 
> Con of that, I suppose, is discoverability, which I was banking on relying on docs for.

Which is why I was originally thinking of an `--api-version` option instead for `install`/`uniinstall`; but that gets complicated by all the OLMv0 options.

Alternatively, we could move OLMv0 to a branch, and move everything to OLMv1. Since that's the direction we're headed.

internal/pkg/action/operator_install.go
joelanford: > The go linter didn't like sets.String.

Something reasonable it didn't like? I haven't seen any issues like that before. Did we add some new linting we really want or did something in 0.26.1 change to trip the linter?

tmshort: The new sets package uses generics; rather than write our own.

After updating go.mod/go.sum, I was getting (I had to revert these changes to demonstrate):
```
tshort@cube:~/git/operator-framework/kubectl-operator (OLM-2869 +)$ make lint
source ./scripts/fetch.sh; fetch golangci-lint 1.50.1 && ./bin/golangci-lint --timeout 3m run
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.
WARN [runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref.
WARN [runner] The linter 'interfacer' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.
internal/pkg/action/operator_install.go:84:74: SA1019: sets.String is deprecated: use generic Set instead. new ways: s1 := Set[string]{} s2 := New[string]() (staticcheck)
func (i *OperatorInstall) possibleInstallModes(watchNamespaces []string) sets.String {
                                                                         ^
internal/pkg/action/operator_install.go:157:99: SA1019: sets.String is deprecated: use generic Set instead. new ways: s1 := Set[string]{} s2 := New[string]() (staticcheck)
func (i OperatorInstall) validateOperatorGroup(og v1.OperatorGroup, operatorInstallModes, desired sets.String) error {
                                                                                                  ^
internal/pkg/action/operator_install.go:200:57: SA1019: sets.String is deprecated: use generic Set instead. new ways: s1 := Set[string]{} s2 := New[string]() (staticcheck)
func (i *OperatorInstall) getTargetNamespaces(supported sets.String) []string {
                                                        ^
make: *** [Makefile:51: lint] Error 1
tshort@cube:~/git/operator-framework/kubectl-operator (OLM-2869 +)$

Comment (tmshort) [reply internal/pkg/action/operator_install.go:14]: It could be due to the updated `k8s.io/apimachinery` package being updated from v0.25.4 to v0.26.1

**Comment by joelanford** [comment]
>  but that gets complicated by all the OLMv0 options.

Yeah exactly. I think we definitely want a separate CLI one way or the other.

> Alternatively, we could move OLMv0 to a branch, and move everything to OLMv1.

We could but I don't think we should do that until OLMv1 GA. There are upstream projects using this with OLMv0 that we would break. (e.g.: https://cert-manager.io/docs/installation/operator-lifecycle-manager/#option-3-manual-install-via-kubectl-operator-plugin)

I think our approaches both do the correct things:
1. Leave the default OLMv1 experience untouched.
2. Ensure we can keep shipping improvements and fixes for OLMv0.
3. Add OLMv1 support such that it is available in releases for those that want to begin experimenting with pre-GA OLMv1.

I suppose we could make a `v1` branch for `v1.y.z[-pre]` tags and essentially maintain `main` for the time being as the v0 branch. That seems overly complicated to me though. And most importantly, it doesn't play well with the krew index which contains a pointer to a single version of the plugin.

IMO putting the OLMv1 stuff in the main branch is a good avenue for letting people know about OLMv1 progress, and maybe getting feedback.

File: internal/pkg/action/operator_install.go (hunk 1 above)
Comment (joelanford) [reply internal/pkg/action/operator_install.go:14]: Yeah, that's it. But it looks like apimachinery has a generic sets.Set type we should switch to.

https://github.com/kubernetes/apimachinery/blob/b4dea92b265131dae13c1e9ebb0a31821bf5daf8/pkg/util/sets/set.go#L24

Comment (joelanford) [reply internal/pkg/action/operator_install.go:14]: Maybe we should handle this in a separate PR though?

**Comment by tmshort** [comment]
ping @awgreene for review?

File: internal/cmd/internal/olmv1/alpha_install.go [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
+	experimentalaction "github.com/operator-framework/kubectl-operator/internal/pkg/experimental/action"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+func NewOperatorInstallCmd(cfg *action.Configuration) *cobra.Command {
+	i := experimentalaction.NewOperatorInstall(cfg)
+	i.Logf = log.Printf
```
dtfranz, line 13: When you set `i.Logf` here does it purely set the log method for the `OperatorInstall` action or could you also use it in this file on lines 23 + 25 i.e. `log.Fatalf(...)` becomes `i.Logf(...)`? (mostly a knowledge question rather than an actual change request)

internal/pkg/experimental/action/operator_uninstall.go
```diff
@@ -0,0 +1,70 @@
...
+		config: cfg,
+		Logf:   func(string, ...interface{}) {},
+	}
+}
+
+func (u *OperatorUninstall) Run(ctx context.Context) error {
+	opKey := types.NamespacedName{Name: u.Package}
+	op := &olmv1.Operator{}
+	op.SetName(opKey.Name)
+	op.SetGroupVersionKind(olmv1.GroupVersion.WithKind("Operator"))
```
dtfranz, line 37: Is this necessary when the object is already explicitly of type `olmv1.Operator`?

internal/pkg/experimental/action/operator_uninstall.go
```diff
@@ -0,0 +1,70 @@
...
+		Name:      obj.GetName(),
+	}
+}
+
+func waitForDeletion(ctx context.Context, cl client.Client, objs ...client.Object) error {
+	for _, obj := range objs {
+		obj := obj
+		lowerKind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
+		key := objectKeyForObject(obj)
+		if err := wait.PollImmediateUntil(250*time.Millisecond, func() (bool, error) {
```
dtfranz, line 58: Could you create a constant for this poll time in the `action` package? Since a couple spots use the same value.

internal/cmd/internal/olmv1/alpha_install.go
tmshort: _Disclaimer:_ this was based on Joe's code; I pulled it in with minimal changes.

AFAICT, setting `i.Logf` is to set the logging destination for the action that is about to happen (i.e. "Logf" = "log function"). It's not really meant to be called as a general logging function. It's clearer to keep lines 23 and 25 as "log.Somethingf", rather than hide it behind `i.Logf`. golang discourages accessors, but this would be a classic case of a set accessor for specify log output.

awgreene (approved):
Looks good to me.

tmshort:
@joelanford @dtfranz is this good?

internal/pkg/experimental/action/operator_uninstall.go
```diff
@@ -0,0 +1,70 @@
...
+
+func objectKeyForObject(obj client.Object) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+}
+
+func waitForDeletion(ctx context.Context, cl client.Client, objs ...client.Object) error {
+	for _, obj := range objs {
```
perdasilva, line 53: nothing that needs to be done rn, but since we are waiting on multiple objects, it could be cool to push each of these waits to its own go routine. Then we can collect the deletion state over each of the objects and feed that back to the user.

perdasilva:
very nice! thank you!!

perdasilva:
looks good to me too ^^ sorry I couldn't approve

tmshort:
> looks good to me too ^^ sorry I couldn't approve

Thanks. @awgreen can likely merge, or @joelanford The [Squash and merge] button is live for me... 🤔

joelanford:
We never directly talked about the envvar enablement mode like in my PR vs the separate subcommand in this one. I'm somewhat partial to the envvar approach but I'm also not going to stand in the way of progress if the consensus is the separate subcommand approach.

tmshort:
The ticket OLM-2869 had it as a an extra keyword as done in this PR.

PR #87, opened 2023-02-03
Update to generic sets

tmshort:
Use the generic sets API (as suggested by go lint) Update go.mod and go.sum

Signed-off-by: Todd Short <tshort@redhat.com>


joelanford (approved):
/lgtm

PR #88, opened 2023-11-16
list-operands: include namespace in error message when package is not found

joelanford:


grokspawn (approved):

grokspawn:
Does this map to an issue?  I didn't see one for unclear error message, and just wanted to check before lgtm-ing.

joelanford:
@joelanford It does not. @dmesser pinged me about it in Slack.

PR #102, opened 2024-01-15
manage tool deps with bingo

joelanford:


m1kola (approved):
I’m hoping that we will get `go tool` with Go 1.22 and bingo will go away.

In the meantime it looks good to me.

/lgtm

PR #115, opened 2024-01-30
add olmv1 subcommands for interfacing with catalog contents

everettraven:
Adds the following subcommands for interacting with catalog contents via catalogd:
- `kubectl operator olmv1 catalog list`
- `kubectl operator olmv1 catalog search`
- `kubectl operator olmv1 catalog inspect`


internal/cmd/internal/olmv1/catalog/alpha_inspect.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	i.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:   "inspect [name]",
+		Short: "inspect catalog objects",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
stevekuznetsov, line 35: Generally speaking, it's best to try to factor CLI commands into a couple of phases - first, gather user input into some `struct{}`, then do any transformations necessary (string parsing, etc), validate the resulting data, and operate on it. This structure helps to make the commands composable but also makes it fairly trivial to use the "meat" of the commands as a library and test them programmatically. See the `kubectl create clusterrole` code as an example of this. 

https://github.com/kubernetes/kubernetes/blob/c4feb19195009e12fa7c49ca1fe91b8898360b0a/staging/src/k8s.io/kubectl/pkg/cmd/create/create_clusterrole.go

internal/pkg/experimental/action/catalog_inspect.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return &CatalogInspectOptions{
+		Catalog: catalog,
+		Schema:  schema,
+		Package: pack,
+		Output:  output,
+	}, nil
+}
+
+func NewCatalogInspect(cfg *action.Configuration, fetcherFunc FetcherFunc, streamerFunc StreamerFunc) (*CatalogInspect, error) {
```
joelanford, line 92: Don't need to return an error in this function definition.

internal/cmd/internal/olmv1/catalog/alpha_inspect.go
```diff
@@ -0,0 +1,72 @@
...
+		func(c *action.Configuration) experimentalaction.CatalogFetcher {
+			return fetcher.New(c.Client)
+		},
+		func(c *action.Configuration) experimentalaction.CatalogContentStreamer {
+			return streamer.New(c.Clientset.CoreV1())
+		},
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
```
joelanford, line 28: If at all possible, it's best to avoid running code that can produce errors as part of the command setup. If an error occurs here, cobra commands are not yet setup, which means:
- Users running an unrelated command will encounter this error
- Users can't run with `--help` and get cobra's usage/help output.

internal/pkg/experimental/action/catalog_inspect.go
```diff
@@ -0,0 +1,160 @@
...
+
+	pack, err := flagSet.GetString(PackageFlag)
+	if err != nil {
+		return nil, fmt.Errorf("parsing package flag: %w", err)
+	}
+
+	output, err := flagSet.GetString(OutputFlag)
+	if err != nil {
+		return nil, fmt.Errorf("parsing output flag: %w", err)
+	}
```
joelanford, line 82: I've never been a fan of this style of flag value fetching, because it introduces a bunch of unnecessary indirection and error checking around the flag name being registered.

I also don't think it makes sense to pollute the `action` package with a bunch of CLI-related features.

My overall suggestion would be:
- Use `flagSet.StringVar*()`-style functions to bind flags directly to existing variables/struct fields
- Move all flag binding code into the cmd package.
- Potentially remove the options struct entirely and put the relevant fields directly on the actions struct.
- Make the `Output` field separate. That is a CLI concern only, right?

internal/pkg/experimental/action/catalog_inspect.go
everettraven: Ah good catch, this was from when I was doing some setup that could error and  i forgot to clean it up 🤦

internal/cmd/internal/olmv1/catalog/alpha_inspect.go
everettraven: Was from https://github.com/operator-framework/kubectl-operator/pull/115#discussion_r1473410860 - should be fixed now

internal/pkg/experimental/action/catalog_inspect.go
everettraven: I'm also not a huge fan of this approach and originally had the `flagSet.StringVar*()` style functions. Will move some stuff around to get rid of this.

everettraven: I just puushed up some changes. I was originally trying to avoid the use of global variables for flags but I am not aware of any other way to achieve this for the time being (I had tried within the function that creates the cobra command but that didn't seem to work, although  I could have been doing something wrong). 

I'm glad you pointed out the flaws with this previous approach. I did keep the options struct as I personally prefer that pattern over requiring it to be on the actions struct.

internal/cmd/internal/olmv1/catalog/alpha_inspect.go
everettraven: I don't disagree. under the hood that is _mostly_ what this is doing. For context underneath the hood it:
- Fetches all catalogs
- Filters all catalogs based on some criteria
- For each catalog (post filtering) stream the contents of the catalog
- For processing, during the streaming of the contents, the contents are appended to the list of catalog objects to return if they meet all the filter criteria based on the provided user input via flags or positional arguments.

As much as possible I was trying to follow the same approach that existed prior to me adding these commands (i.e: https://github.com/operator-framework/kubectl-operator/blob/a6eb73a477ceb2295eb9422ccc7b1e7c23db4756/internal/cmd/operator_install.go#L16-L36)

I _personally_ think there is some restructuring of this repo that could be done to make it a bit more intuitive on how to add subcommands and extend the functionality of this plugin but that is out of scope for this PR

stevekuznetsov: I'm only speaking to the implementations that are new here - for instance, there's no way to test the post-processing after `Run()` here nor are any invariants validated about `experimentalaction.CatalogInspectOptions` - if you can factor this to collapse the `*cobra.Command` run func to nothing but passing opts elsewhere, which you are super close to, I think you'd find it easier to validate your commands programmatically.

Issue #86, opened 2023-02-02
Define strategy for migration from OLMv0 to OLMv1


joelanford:
As of #85, we have begun to add support for the OLMv1 APIs into this plugin via a new subcommand `kubectl operator olmv1`. So the current state of the world is:
- OLMv1 stuff at `kubectl operator olmv1`
- OLMv0 stuff at `kubectl operator *` (except `olmv1`)

Long term, I think the goal is:
- OLMv1 stuff at `kubectl operator`
- OLMv0 stuff no longer present.

How do we get from A to B?

## Proposal 1

In #84, I proposed the use of an environment variable (I chose `EXPERIMENTAL_USE_OLMV1_APIS`, but we could pick something else), which would act as a switch controlling the entirety of the `kubectl operator` functionality:
- `EXPERIMENTAL_USE_OLMV1_APIS=on`: `kubectl operator` is OLMv1
- Anything else: `kubectl operator` is OLMv0

My thought there was that we could perform the migration like this:
- T1. OLMv0 support only
- T2. Default: OLMv0, require EXPERIMENTAL_USE_OLMV1_APIS=on for OLMv1
- T3. Default: OLMv1, require EXPERIMENTAL_USE_OLMV1_APIS=off for OLMv0
- T4. OLMv1 support only

In this proposed model, OLMv1 users can essentially opt-in early (at T2) and then not have to worry about subcommand changes later. Also in this model, OLMv0 users can explicitly opt-out early to avoid a breakage at T3. This gives both sets of users more time to adjust to incoming changes.

## Proposal 2

Another option is to perform this same transition, but using subcommand renaming:
- T1. OLMv0 support only at `kubectl operator`
- T2. OLMv0 support at `kubectl operator`, OLMv1 support at `kubectl operator olmv1`
- T3. OLMv1 support at `kubectl operator`, OLMv0 support at `kubectl operator olmv0`
- T4. OLMv1 support only at `kubectl operator`

## Proposal 3

Another option is a variation on Proposal 2, except we skip T3 and do a hard cutover:
- T1. OLMv0 support only at `kubectl operator`
- T2. OLMv0 support at `kubectl operator`, OLMv1 support at `kubectl operator olmv1`
- T3. OLMv1 support only at `kubectl operator`

## My two cents

[...]

PR #218, opened 2025-02-19
✨  Add commands to list and get olmv1 operators and catalogs

azych:
This adds `get` command that enables getting/listing olmv1 `operators`(`ClusterExtensions`) and `catalogs` (`ClusterCatalogs`) via the CLI.

Syntax follows the `get` behavior of `kubectl`:
`olmv1 get operator(s)` - list all operators
`olmv1 get operator(s) <name>` - get details of a single operator
`olmv1 get catalog(s)` - list all catalogs
`olmv1 get catalog(s) <name>` - get details of a single catalog

Commands return data in the following format similar to `kubectl` output, with catalogs being sorted by availability, priority and name; operators being sorted by name:
```bash
➜  go run *.go olmv1 get operator
NAME       INSTALLED BUNDLE            VERSION  SOURCE TYPE  INSTALLED  PROGRESSING  AGE
argocd     argocd-operator.v0.6.0      0.6.0    Catalog      True       True         5h44m
zookeeper  zookeeper-operator.v0.17.8  0.17.8   Catalog      True       True         24s
➜  go run *.go olmv1 get operators
NAME       INSTALLED BUNDLE            VERSION  SOURCE TYPE  INSTALLED  PROGRESSING  AGE
argocd     argocd-operator.v0.6.0      0.6.0    Catalog      True       True         5h44m
zookeeper  zookeeper-operator.v0.17.8  0.17.8   Catalog      True       True         30s
➜  go run *.go olmv1 get operator zookeeper
NAME       INSTALLED BUNDLE            VERSION  SOURCE TYPE  INSTALLED  PROGRESSING  AGE
zookeeper  zookeeper-operator.v0.17.8  0.17.8   Catalog      True       True         37s
➜  go run *.go olmv1 get operators zookeeper
NAME       INSTALLED BUNDLE            VERSION  SOURCE TYPE  INSTALLED  PROGRESSING  AGE
zookeeper  zookeeper-operator.v0.17.8  0.17.8   Catalog      True       True         41s
➜  go run *.go olmv1 get catalog
NAME              AVAILABILITY  PRIORITY  LASTUNPACKED  SERVING  AGE
notoperatorhubio  Available     0         13s           True     20s
operatorhubio     Available     0         103m          True     5h46m
➜  go run *.go olmv1 get catalogs
NAME              AVAILABILITY  PRIORITY  LASTUNPACKED  SERVING  AGE
notoperatorhubio  Available     0         19s           True     26s
operatorhubio     Available     0         103m          True     5h46m
➜  go run *.go olmv1 get catalog operatorhubio
NAME           AVAILABILITY  PRIORITY  LASTUNPACKED  SERVING  AGE
operatorhubio  Available     0         103m          True     5h47m
➜  go run *.go olmv1 get catalogs operatorhubio
NAME           AVAILABILITY  PRIORITY  LASTUNPACKED  SERVING  AGE
operatorhubio  Available     0         103m          True     5h47m
```

closes https://github.com/operator-framework/operator-controller/issues/1767

cc @ankitathomas

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,12 +5,14 @@ go 1.23.0
 toolchain go1.23.4
 
 require (
+	github.com/blang/semver/v4 v4.0.0
 	github.com/containerd/containerd v1.7.25
 	github.com/containerd/platforms v0.2.1
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.36.2
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/operator-framework/api v0.29.0
+	github.com/operator-framework/catalogd v1.1.0
 	github.com/operator-framework/operator-controller v1.1.0
 	github.com/operator-framework/operator-lifecycle-manager v0.23.1
 	github.com/operator-framework/operator-registry v1.50.0
@@ -32,7 +34,6 @@ require (
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/Microsoft/hcsshim v0.12.9 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
-	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/containerd/cgroups/v3 v3.0.3 // indirect
 	github.com/containerd/containerd/api v1.8.0 // indirect
 	github.com/containerd/continuity v0.4.4 // indirect
--- a/internal/cmd/internal/olmv1/catalog_installed_get.go
+++ b/internal/cmd/internal/olmv1/catalog_installed_get.go
@@ -1,0 +1,38 @@
+package olmv1
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
+	v1action "github.com/operator-framework/kubectl-operator/internal/pkg/v1/action"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+// NewCatalogInstalledGetCmd handles get commands in the form of:
+// catalog(s) [catalog_name] - this will either list all the installed operators
+// if no catalog_name has been provided or display the details of the specific
+// one otherwise
+func NewCatalogInstalledGetCmd(cfg *action.Configuration) *cobra.Command {
+	i := v1action.NewCatalogInstalledGet(cfg)
+	i.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:     "catalog [catalog_name]",
+		Aliases: []string{"catalogs"},
+		Args:    cobra.RangeArgs(0, 1),
+		Short:   "Display one or many installed catalogs",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) == 1 {
+				i.CatalogName = args[0]
+			}
+			installedCatalogs, err := i.Run(cmd.Context())
+			if err != nil {
+				log.Fatalf("failed getting installed catalog(s): %v", err)
+			}
+
+			printFormattedCatalogs(installedCatalogs...)
+		},
+	}
+
+	return cmd
+}
--- a/internal/cmd/internal/olmv1/printing.go
+++ b/internal/cmd/internal/olmv1/printing.go
@@ -1,0 +1,90 @@
+package olmv1
+
+import (
+	"cmp"
+	"fmt"
+	"os"
+	"slices"
+	"text/tabwriter"
+	"time"
+
+	"github.com/blang/semver/v4"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/duration"
+
+	catalogdv1 "github.com/operator-framework/catalogd/api/v1"
+	olmv1 "github.com/operator-framework/operator-controller/api/v1"
+)
+
+func printFormattedOperators(extensions ...olmv1.ClusterExtension) {
+	tw := tabwriter.NewWriter(os.Stdout, 3, 4, 2, ' ', 0)
+	_, _ = fmt.Fprint(tw, "NAME\tINSTALLED BUNDLE\tVERSION\tSOURCE TYPE\tINSTALLED\tPROGRESSING\tAGE\n")
+
+	sortOperators(extensions)
+	for _, ext := range extensions {
+		age := time.Since(ext.CreationTimestamp.Time)
+		_, _ = fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
+			ext.Name,
+			ext.Status.Install.Bundle.Name,
+			ext.Status.Install.Bundle.Version,
+			ext.Spec.Source.SourceType,
+			status(ext.Status.Conditions, olmv1.TypeInstalled),
+			status(ext.Status.Conditions, olmv1.TypeProgressing),
+			duration.HumanDuration(age),
+		)
+	}
+	_ = tw.Flush()
+}
+
+func printFormattedCatalogs(catalogs ...catalogdv1.ClusterCatalog) {
+	tw := tabwriter.NewWriter(os.Stdout, 3, 4, 2, ' ', 0)
+	_, _ = fmt.Fprint(tw, "NAME\tAVAILABILITY\tPRIORITY\tLASTUNPACKED\tSERVING\tAGE\n")
+
+	sortCatalogs(catalogs)
+	for _, cat := range catalogs {
+		age := time.Since(cat.CreationTimestamp.Time)
+		lastUnpacked := time.Since(cat.Status.LastUnpacked.Time)
+		_, _ = fmt.Fprintf(tw, "%s\t%s\t%d\t%s\t%s\t%s\n",
+			cat.Name,
+			string(cat.Spec.AvailabilityMode),
+			cat.Spec.Priority,
+			duration.HumanDuration(lastUnpacked),
+			status(cat.Status.Conditions, catalogdv1.TypeServing),
+			duration.HumanDuration(age),
+		)
+	}
+	_ = tw.Flush()
+}
+
+// sortOperators sorts operators in place and uses the following sorting order:
+// name (asc), version (desc)
+func sortOperators(extensions []olmv1.ClusterExtension) {
+	slices.SortFunc(extensions, func(a, b olmv1.ClusterExtension) int {
+		return cmp.Or(
+			cmp.Compare(a.Name, b.Name),
+			-semver.MustParse(a.Status.Install.Bundle.Version).Compare(semver.MustParse(b.Status.Install.Bundle.Version)),
+		)
+	})
+}
+
+// sortCatalogs sorts catalogs in place and uses the following sorting order:
+// availability (asc), priority (desc), name (asc)
+func sortCatalogs(catalogs []catalogdv1.ClusterCatalog) {
+	slices.SortFunc(catalogs, func(a, b catalogdv1.ClusterCatalog) int {
+		return cmp.Or(
+			cmp.Compare(a.Spec.AvailabilityMode, b.Spec.AvailabilityMode),
+			-cmp.Compare(a.Spec.Priority, b.Spec.Priority),
+			cmp.Compare(a.Name, b.Name),
+		)
+	})
+}
+
+func status(conditions []metav1.Condition, typ string) string {
+	for _, condition := range conditions {
+		if condition.Type == typ {
+			return string(condition.Status)
+		}
+	}
+
+	return "Unknown"
+}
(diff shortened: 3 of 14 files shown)


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/onsi/gomega v1.36.2
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/operator-framework/api v0.29.0
+	github.com/operator-framework/catalogd v1.1.0
```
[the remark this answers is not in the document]
joelanford: I'd go ahead and bump to the latest (non-rc) of operator-controller.

joelanford:
get and list should not be separate subcommands. With kubectl, one can list by simply leaving the name argument out of the command args.

So
- `kubectl operator get catalog(s)` -> lists the ClusterCatalogs
- `kubectl operator get catalog(s) my-catalog -> gets the ClusterCatalog called `my-catalog`

internal/cmd/internal/olmv1/printing.go
```diff
@@ -0,0 +1,58 @@
...
+			duration.HumanDuration(age),
+		)
+	}
+	_ = tw.Flush()
+}
+
+func printFormattedCatalogs(catalogs ...catalogdv1.ClusterCatalog) {
+	tw := tabwriter.NewWriter(os.Stdout, 3, 4, 2, ' ', 0)
+	_, _ = fmt.Fprint(tw, "NAME\tLASTUNPACKED\tSERVING\tAGE\n")
+	for _, cat := range catalogs {
```
joelanford, line 37: Now or in a follow-up, I don't have an opinion which:

I would suggest sorting these:
- First by enabled/disabled
- Next by priority
- Next by name

joelanford: And we should include columns for enabled/disable and priority.

azych:
> get and list should not be separate subcommands. With kubectl, one can list by simply leaving the name argument out of the command args.
> 
> So
> 
>     * `kubectl operator get catalog(s)` -> lists the ClusterCatalogs
> 
>     * `kubectl operator get catalog(s) my-catalog -> gets the ClusterCatalog called `my-catalog`

I was going by the separation that already exists in olmv0 commands but this approach would be my preference as well, so happy to make the changes

internal/cmd/internal/olmv1/operator_installed_list.go
```diff
@@ -0,0 +1,34 @@
...
+	v1action "github.com/operator-framework/kubectl-operator/internal/pkg/v1/action"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+func NewOperatorInstalledListCmd(cfg *action.Configuration) *cobra.Command {
+	i := v1action.NewOperatorInstalledList(cfg)
+	i.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:   "operators",
```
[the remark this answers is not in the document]
azych: I agree about this being highly subjective, for me I'd generally prefer predictable structure because with that I can follow the logic and probably will get the desired effect I'm assuming I'd get.
At the same time I see your point about additional typing and since our base is `operator` it wouldn't be complete memorization ;)

PR-wise I don't have strong feelings here and I'll be ok with either approach

internal/cmd/internal/olmv1/printing.go
```diff
@@ -0,0 +1,58 @@
...
+	tw := tabwriter.NewWriter(os.Stdout, 3, 4, 2, ' ', 0)
+	_, _ = fmt.Fprint(tw, "NAME\tINSTALLED BUNDLE\tVERSION\tSOURCE TYPE\tINSTALLED\tPROGRESSING\tAGE\n")
+	for _, ext := range extensions {
+		age := time.Since(ext.CreationTimestamp.Time)
+		_, _ = fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
+			ext.Name,
+			ext.Status.Install.Bundle.Name,
+			ext.Status.Install.Bundle.Version,
+			ext.Spec.Source.SourceType,
+			status(ext.Status.Conditions, olmv1.TypeInstalled),
```
bentito, line 26: when you get to unit tests at some point, probably should have one (+?)  showing we don't list olmv0 resources. I think `kubectl-operator` doesn't know what might be on a cluster, could have v0 and v1 operators installed, certainly catalog can have both.

azych: does `olmv0` use the same api version and kind as `olmv1` - ie. `olm.operatorframework.io/v1` and `ClusterExtension`?

internal/cmd/olmv1.go
```diff
@@ -14,9 +14,33 @@ func newOlmV1Cmd(cfg *action.Configuration) *cobra.Command {
 		Long:  "Manage operators via OLMv1 in a cluster from the command line.",
 	}
 
+	// use get <resource_type>
```
anik120, line 17: Intentional comment?

internal/cmd/internal/olmv1/operator_installed_list.go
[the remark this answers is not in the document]
azych: BTW. our base is actually `operator olmv1 <command>` so with that I think it would be preferable to leave the structure in

internal/cmd/internal/olmv1/printing.go
bentito: yeah we did make the break to new resource kinds, so I guess it's safe

go.mod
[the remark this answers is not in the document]
azych: since 1.1.0 is the latest non-rc operator-controller version, we'll have to follow up when we have a new non-rc with monorepo changes ready

internal/cmd/internal/olmv1/printing.go
```diff
@@ -0,0 +1,78 @@
...
+	olmv1 "github.com/operator-framework/operator-controller/api/v1"
+)
+
+func printFormattedOperators(extensions ...olmv1.ClusterExtension) {
+	tw := tabwriter.NewWriter(os.Stdout, 3, 4, 2, ' ', 0)
+	_, _ = fmt.Fprint(tw, "NAME\tINSTALLED BUNDLE\tVERSION\tSOURCE TYPE\tINSTALLED\tPROGRESSING\tAGE\n")
+
+	// sort by name
+	slices.SortFunc(extensions, func(a, b olmv1.ClusterExtension) int {
+		return cmp.Compare(a.Name, b.Name)
```
grokspawn, line 24: It's been a little while, but I'm pretty sure we use blang.semver for bundle versions, and masterminds.semver for the ranges (but I'm also pretty sure that the regex is functionally different _only_ for ranges and wouldn't otherwise matter).

```suggestion
	slices.SortFunc(extensions, func(a, b olmv1.ClusterExtension) int {
		return cmp.OR(
		    cmp.Compare(a.Name, b.Name)
		    bsemver.MustParse(a.Status.Install.Bundle.Version).Compare(bsemver.MustParse(b.Status.Install.Bundle.Version))
		    )
```
... or similar

perdasilva:
I don't think it's necessary to get this PR down. But, if it's easy enough, it would be cool to have an asciinema demo attached to the PR description (maybe even the script checked-in with the PR =D)

internal/cmd/internal/olmv1/operator_installed_list.go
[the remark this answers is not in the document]
joelanford: > BTW. our base is actually operator olmv1 <command> so with that I think it would be preferable to leave the structure in

Yeah, I realize that is our current base. But that is leftover from a _very_ early effort to add OLMv1 support when OLMv1 was an alpha level project.

Whenever we feel like we are ready, I would like for us to make the OLMv1 entrypoints the "main" ones. And either relegate the OLMv0 subcommands behind some opt-in knob or just remove them.

internal/cmd/internal/olmv1/printing.go
joelanford: > I think kubectl-operator doesn't know what might be on a cluster, could have v0 and v1 operators installed, certainly catalog can have both.

But it should not try to detect and change its operating mode automatically.

At the risk of being too opinionated, let me try to be unequivocal. As soon as sufficient functionality exists in kubectl-operator to support OLMv1, we should either remove OLMv0 support or put it somewhere that requires a user to specifically decide to use it.

internal/cmd/internal/olmv1/operator_installed_list.go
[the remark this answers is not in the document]
joelanford: Relevant issue: https://github.com/operator-framework/kubectl-operator/issues/86

internal/cmd/internal/olmv1/catalog_installed_get.go
```diff
@@ -0,0 +1,38 @@
...
+// NewCatalogInstalledGetCmd handles get commands in the form of:
+// catalog(s) [operatorName] - this will either list all the installed operators
+// if no operatorName has been provided or display the details of the specific
+// one otherwise
+func NewCatalogInstalledGetCmd(cfg *action.Configuration) *cobra.Command {
+	i := v1action.NewCatalogInstalledGet(cfg)
+	i.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:     "catalog",
```
joelanford, line 20: I don't know if we're consistent with this suggestion now, but it would be nice to include the optional arg in the `Use` string. E.g. something like:

```suggestion
		Use:     "catalog [catalog_name]",
```

If we're not already consistently doing this, capturing a follow-up issue to improve our usage text would suffice.

internal/cmd/internal/olmv1/operator_installed_list.go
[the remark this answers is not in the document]
azych: I agree, there is a separate effort to deprecate olmv0 which I think this belongs to, so I'll resolve this here. 

It also shouldn't be difficult to hook up logic of the new commands that we're adding here and will be following up with additional ones to a different upper level command according to the decisions we make when we discuss approach to olmv0 deprecation.

internal/cmd/internal/olmv1/printing.go
azych: AFAIR it might be the other way around or at least we should be using blang for ranges, but I agree that it shouldn't which lib we use to parse the version here.

I went with blang because it was already an indirect dependency because of olmv0. If someone has an opinion why blang shouldn't be used here, we can switch when we deprecate olmv0.

internal/cmd/internal/olmv1/catalog_installed_get.go
azych: thanks, we were actually doing that consistently both in our old alpha v1 commands as well as across olmv0 commands in general and I extended the usage to match

internal/pkg/v1/action/catalog_installed_get.go
```diff
@@ -0,0 +1,47 @@
...
+
+func NewCatalogInstalledGet(cfg *action.Configuration) *CatalogInstalledGet {
+	return &CatalogInstalledGet{
+		config: cfg,
+		Logf:   func(string, ...interface{}) {},
+	}
+}
+
+func (i *CatalogInstalledGet) Run(ctx context.Context) ([]olmv1catalogd.ClusterCatalog, error) {
+	// get
```
perdasilva, line 29:
```suggestion
```

perdasilva: or maybe something less terse?

internal/pkg/v1/action/catalog_installed_get.go
```diff
@@ -0,0 +1,47 @@
...
+		opKey := types.NamespacedName{Name: i.CatalogName}
+		err := i.config.Client.Get(ctx, opKey, &result)
+		if err != nil {
+			return nil, err
+		}
+
+		return []olmv1catalogd.ClusterCatalog{result}, nil
+	}
+
+	// list
```
perdasilva, line 42: same as above

internal/pkg/v1/action/catalog_installed_get_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	olmv1catalogd "github.com/operator-framework/catalogd/api/v1"
+
+	internalaction "github.com/operator-framework/kubectl-operator/internal/pkg/v1/action"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+var _ = Describe("CatalogInstalledGet", func() {
```
perdasilva, line 21: is it worth exercising the sort order?

perdasilva (approved):
Some small nits around comments and wondering whether we ought to also test the sort order. But, not super worried about it. Leave it to you to decide ^^ otherwise - looks nice! very concise and easy to read. Thank you!

internal/pkg/v1/action/catalog_installed_get.go
azych: seems fine to me, so I'll leave it in

internal/pkg/v1/action/catalog_installed_get_test.go
azych: I asked the same question and originally thought that it be too much like testing the built-in cmp functionalities, but I guess an additional test case won't hurt

grokspawn:
Great job @azych!

PR #219, opened 2025-02-26
✨  Add command to delete an existing olmv1 catalog

azych:
This adds `delete` command that enables deleting either a single or all existing catalogs (`ClusterCatalog`) from the cluster.

Signature:
```bash
Delete either a single or all of the existing catalogs

Usage:
  operator olmv1 delete catalog [catalog_name] [flags]

Aliases:
  catalog, catalogs [catalog_name]

Flags:
      --all    delete all catalogs
  -h, --help   help for catalog
```

Example walkthrough:
```bash
➜  kubectl get clustercatalogs
NAME            LASTUNPACKED   SERVING   AGE
operatorhubio   5m41s          True      5m51s
➜ go run *.go olmv1 delete catalog unknown
failed to delete catalog "unknown": clustercatalogs.olm.operatorframework.io "unknown" not found
exit status 1
➜ go run *.go olmv1 delete catalog operatorhubio
catalog "operatorhubio" deleted
➜ kubectl get clustercatalogs
No resources found

➜ kubectl get clustercatalogs
NAME             LASTUNPACKED   SERVING   AGE
operatorhubio    2m4s           True      2m11s
operatorhubio2   5s             True      18s
➜ go run *.go olmv1 delete catalog --all
catalog "operatorhubio" deleted
catalog "operatorhubio2" deleted
➜ kubectl get clustercatalogs
No resources found

➜ kubectl get clustercatalogs
NAME             LASTUNPACKED   SERVING   AGE
operatorhubio    55s            True      62s
operatorhubio2   45s            True      59s
➜ kubectl scale deploy catalogd-controller-manager operator-controller-controller-manager --replicas=0 -n olmv1-system
➜ go run *.go olmv1 delete catalog --all
failed deleting catalogs: failed deleting catalog "operatorhubio": waiting for deletion: context deadline exceeded
failed deleting catalog "operatorhubio2": client rate limiter Wait returned an error: context deadline exceeded
exit status 1

➜ kubectl get clustercatalogs
NAME             LASTUNPACKED   SERVING   AGE
operatorhubio2   23s            True      30s
➜ go run *.go olmv1 delete catalog operatorhubio2 --all
failed to delete catalog "operatorhubio2": name cannot be provided when a selector is specified
exit status 1
➜ go run *.go olmv1 delete catalog --all operatorhubio2
failed to delete catalog "operatorhubio2": name cannot be provided when a selector is specified
exit status 1
```

Additionally, does a light refactor of the old helpers, especially the `waitForDeletion`, which does not accept variable number of input objects anymore.
Benefits of this are:
* callers have more control over deletion and can decide how they want to run it - async/sync for example
* de-duplicated error messages - no need to mention details of the object within the helper
* no need for extending the deleting object with GVK info just for the sake of helper and its logging  

~There is some overlap with https://github.com/operator-framework/kubectl-operator/pull/218 - ie. test helpers, which will be cleaned up once that has been merged~ - done

Part of https://github.com/operator-framework/operator-controller/pull/1815

### DIFF
--- a/internal/cmd/internal/olmv1/catalog_delete.go
+++ b/internal/cmd/internal/olmv1/catalog_delete.go
@@ -1,0 +1,50 @@
+package olmv1
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/operator-framework/kubectl-operator/internal/cmd/internal/log"
+	v1action "github.com/operator-framework/kubectl-operator/internal/pkg/v1/action"
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+// NewCatalogDeleteCmd allows deleting either a single or all
+// existing catalogs
+func NewCatalogDeleteCmd(cfg *action.Configuration) *cobra.Command {
+	d := v1action.NewCatalogDelete(cfg)
+	d.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:     "catalog [catalog_name]",
+		Aliases: []string{"catalogs [catalog_name]"},
+		Args:    cobra.RangeArgs(0, 1),
+		Short:   "Delete either a single or all of the existing catalogs",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) == 0 {
+				catalogs, err := d.Run(cmd.Context())
+				if err != nil {
+					log.Fatalf("failed deleting catalogs: %v", err)
+				}
+				for _, catalog := range catalogs {
+					log.Printf("catalog %q deleted", catalog)
+				}
+
+				return
+			}
+
+			d.CatalogName = args[0]
+			if _, err := d.Run(cmd.Context()); err != nil {
+				log.Fatalf("failed to delete catalog %q: %v", d.CatalogName, err)
+			}
+			log.Printf("catalog %q deleted", d.CatalogName)
+		},
+	}
+	bindCatalogDeleteFlags(cmd.Flags(), d)
+
+	return cmd
+}
+
+func bindCatalogDeleteFlags(fs *pflag.FlagSet, d *v1action.CatalogDelete) {
+	fs.BoolVar(&d.DeleteAll, "all", false, "delete all catalogs")
+}
--- a/internal/pkg/v1/action/catalog_delete.go
+++ b/internal/pkg/v1/action/catalog_delete.go
@@ -1,0 +1,69 @@
+package action
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	olmv1catalogd "github.com/operator-framework/catalogd/api/v1"
+
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+type CatalogDelete struct {
+	config      *action.Configuration
+	CatalogName string
+	DeleteAll   bool
+
+	Logf func(string, ...interface{})
+}
+
+func NewCatalogDelete(cfg *action.Configuration) *CatalogDelete {
+	return &CatalogDelete{
+		config: cfg,
+		Logf:   func(string, ...interface{}) {},
+	}
+}
+
+func (cd *CatalogDelete) Run(ctx context.Context) ([]string, error) {
+	// validate
+	if cd.DeleteAll && cd.CatalogName != "" {
+		return nil, errNameAndSelector
+	}
+
+	// delete single, specified catalog
+	if !cd.DeleteAll {
+		return nil, cd.deleteCatalog(ctx, cd.CatalogName)
+	}
+
+	// delete all existing catalogs
+	var catatalogList olmv1catalogd.ClusterCatalogList
+	if err := cd.config.Client.List(ctx, &catatalogList); err != nil {
+		return nil, err
+	}
+	if len(catatalogList.Items) == 0 {
+		return nil, errNoResourcesFound
+	}
+
+	errs := make([]error, 0, len(catatalogList.Items))
+	names := make([]string, 0, len(catatalogList.Items))
+	for _, catalog := range catatalogList.Items {
+		names = append(names, catalog.Name)
+		if err := cd.deleteCatalog(ctx, catalog.Name); err != nil {
+			errs = append(errs, fmt.Errorf("failed deleting catalog %q: %w", catalog.Name, err))
+		}
+	}
+
+	return names, errors.Join(errs...)
+}
+
+func (cd *CatalogDelete) deleteCatalog(ctx context.Context, name string) error {
+	op := &olmv1catalogd.ClusterCatalog{}
+	op.SetName(name)
+
+	if err := cd.config.Client.Delete(ctx, op); err != nil {
+		return err
+	}
+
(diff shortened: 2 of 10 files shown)


internal/cmd/internal/olmv1/catalog_delete.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+// NewCatalogDeleteCmd allows deleting a specified, existing catalog
+func NewCatalogDeleteCmd(cfg *action.Configuration) *cobra.Command {
+	i := v1action.NewCatalogDelete(cfg)
+	i.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:     "catalog <catalog_name>",
```
ankitathomas, line 17: It'd be useful to add a label filter or an `--all` option to use in case a catalog name isn't specified

ankitathomas (approved):
/lgtm

internal/cmd/internal/olmv1/catalog_delete.go
azych: I agree, `--all` sounds good.
Regarding possible `selector` I think it's best if it's introduced separately, since it would make sense for multiple commands

azych: I added `--all` with deletions waits being done synchronously and using the same command-level timeout. I think that's fine now, but I'll add an issue to track potentially adding an async option to speed things up. This also kind of depends on how to handle waiting - https://github.com/operator-framework/kubectl-operator/pull/220#discussion_r1973356442

internal/pkg/v1/action/catalog_delete.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	errs := make([]error, 0, len(catsList.Items))
+	names := make([]string, 0, len(catsList.Items))
+	for _, cat := range catsList.Items {
+		names = append(names, cat.Name)
+		if err := cd.deleteCatalog(ctx, cat.Name); err != nil {
+			errs = append(errs, fmt.Errorf("failed deleting catalog %q: %w", cat.Name, err))
+		}
+	}
```
ankitathomas, line 55: I think you can avoid listing separately with the controllerruntime client's `DeleteAllOf`.

LalatenduMohanty:
Currently the command is 
```
$ kubectl operator olmv1 delete  catalog operatorhubio
catalog "operatorhubio" deleted
```
And I think is too long. I would love it to be as `$ kubectl operator olmv1 delete --catalog <name of the catalog or catalogs>`
```
$ kubectl operator olmv1 delete --catalog operatorhubio
```
WDYT?

LalatenduMohanty:
I am wondering if we should add an flag `--all` to `kubectl operator olmv1 delete catalog` to delete all the catalogs. We can do it in another PR if we think it will help, but it is not very important or urgent right now. 
```
$ kubectl operator olmv1 delete catalog
failed deleting catalogs: resource name may not be empty
```

internal/pkg/v1/action/catalog_delete.go
azych: I specifically avoided using `DeleteAllOf` and went with list -> delete, so that we:
* know what catalogs we are going to delete
* can wait for each one separately
* can return a more detailed/helpful message back to the user (either which catalogs were deleted or which failed deletion and why each of those failed).

azych:
> Currently the command is
> 
> ```
> $ kubectl operator olmv1 delete  catalog operatorhubio
> catalog "operatorhubio" deleted
> ```
> 
> And I think is too long. I would love it to be as `$ kubectl operator olmv1 delete --catalog <name of the catalog or catalogs>`
> 
> ```
> $ kubectl operator olmv1 delete --catalog operatorhubio
> ```
> 
> WDYT?

Your idea to change the type to be a flag on the command makes the overall command longer.

Overall I think we should continue with the syntax we agreed on and used so far. If you feel strongly about this, maybe olmv0 deprecation might be a good occasion to propose a new syntax idea? As part of olmv0 deprecation we might modify the current command 'prefix' - `operator olmv1'.

Hope that's fair.

internal/cmd/internal/olmv1/catalog_delete.go
```diff
@@ -0,0 +1,50 @@
...
+	d.Logf = log.Printf
+
+	cmd := &cobra.Command{
+		Use:     "catalog [catalog_name]",
+		Aliases: []string{"catalogs [catalog_name]"},
+		Args:    cobra.RangeArgs(0, 1),
+		Short:   "Delete either a single or all of the existing catalogs",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) == 0 {
+				cats, err := d.Run(cmd.Context())
```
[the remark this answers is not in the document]
azych: I agree, let's try to keep abbreviations like that mostly in small helpers and tests

internal/pkg/v1/action/catalog_delete.go
```diff
@@ -0,0 +1,69 @@
...
+
+	olmv1catalogd "github.com/operator-framework/catalogd/api/v1"
+
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+type CatalogDelete struct {
+	config      *action.Configuration
+	CatalogName string
+	DeleteAll   bool
```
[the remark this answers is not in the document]
azych: it is for deleting all the catalogs though, so I'll leave it as-is

internal/pkg/v1/action/catalog_delete.go
ankitathomas: That's a good point, it's certainly helpful to have detailed error messages for failing to delete specific catalogs. 

I am a little worried about sending too many API requests on mass deletes - but hopefully mass deleting catalogs shouldn't be a common operation. `DeleteAllOf` also comes with support for label selectors which we may need to support in future, though we can get pretty much the same thing with the initial List.

ankitathomas:
> > I am wondering if we should add an flag `--all` to `kubectl operator olmv1 delete catalog` to delete all the catalogs. We can do it in another PR if we think it will help, but it is not very important or urgent right now.
> > ```
> > $ kubectl operator olmv1 delete catalog
> > failed deleting catalogs: resource name may not be empty
> > ```
> 
> This PR already includes `--all`

@LalatenduMohanty do you mean treating `olmv1 delete catalog --all` as successful when there are no clustercatalogs to delete?

internal/pkg/v1/action/catalog_delete.go
azych: I agree, this is a concern. We probably might also at some point consider what level of performance are we expecting from the CLI.

Specifically regarding `DeleteAllOf`, if separate deletes turn out to be problematic, we can probably make a tradeoff between being performant and being specific in output to users. As a middle way, we could still leave the initial `List()` call and just combine it with `DeleteAllOf` like you suggested.
I'll look into the behavior more in depth and if we won't be missing much then potentially adjust it here or in a later PR.

LalatenduMohanty:
> This PR already includes --all

PR #224, opened 2025-03-05
🌱  Bump operator-controller; remove operator-framework/catalogd from deps

azych:
* bumps operator-controller to non-rc monorepo version (1.2.0)
* removes operator-framework/catalogd dependency as it's no longer needed
* refactors references to operator-framework/catalogd API types to point to operator-controller


go.mod
```diff
@@ -1,29 +1,28 @@
 module github.com/operator-framework/kubectl-operator
 
-go 1.23.0
+go 1.23.4
 
-toolchain go1.23.4
+toolchain go1.23.6
```
camilamacedo86, line 5: I think we need to keep it as it is all projects to make the life easier
go 1.23.0 and go1.23.4

azych: Looks we don't specify it in operator-controller - maybe we don't need to do that here as well?

LalatenduMohanty: +1 to remove it. cc @tmshort .

LalatenduMohanty:
Looks good  to me. Once we fix https://github.com/operator-framework/kubectl-operator/pull/224#discussion_r1981105192 we should be able to merge it.

go.mod
tmshort: We want to avoid anything above 1.23.4, try explicitly using 1.23.4, and re-tidy this go.mod

tmshort: (i.e. download go 1.23.4, and use that for any `go` operations)

azych: did `go get go@1.23.4`, reverted toolchain change and `go tidied` - the result is go 1.23.4 with no explicit toolchain line (probably as per `For example, a go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line.` - https://go.dev/doc/toolchain#config)

we should be good to go 🎊

go.mod
```diff
@@ -1,29 +1,26 @@
 module github.com/operator-framework/kubectl-operator
 
-go 1.23.0
-
-toolchain go1.23.4
+go 1.23.4
```
camilamacedo86, line 3: Do we need to change it?
Can we not just leave it as it is?

camilamacedo86, line 3: Since it is as we do in oper-controller: https://github.com/operator-framework/operator-controller/blob/main/go.mod#L3

I am fine with the change 👍

camilamacedo86 (approved):
/lgtm

PR #226, opened 2025-03-07
fix empty client in action config

ankitathomas:
Fixing the panic because of the empty client in action config
```
$ kubectl operator olmv1 create catalog operatorhub quay.io/operatorhubio/catalog:latest
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x184c59c]

goroutine 1 [running]:
github.com/operator-framework/kubectl-operator/internal/pkg/v1/action.(*CatalogCreate).Run(0xc0001126c0, {0x1f6f468, 0xc0002b19d0})
	kubectl-operator/internal/pkg/v1/action/catalog_create.go:41 +0x1dc
github.com/operator-framework/kubectl-operator/internal/cmd/internal/olmv1.NewCatalogCreateCmd.func1(0xc000374c08?, {0xc00030ed60?, 0x4?, 0x1c59745?})
	kubectl-operator/internal/cmd/internal/olmv1/catalog_create.go:28 +0xaf
github.com/spf13/cobra.(*Command).execute(0xc000374c08, {0xc00030ed20, 0x2, 0x2})
	/home/lmohanty/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989 +0xa91
github.com/spf13/cobra.(*Command).ExecuteC(0xc00040e908)
	/home/lmohanty/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
	/home/lmohanty/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
github.com/operator-framework/kubectl-operator/internal/cmd.Execute()
	kubectl-operator/internal/cmd/root.go:14 +0x18
main.main()
	kubectl-operator/main.go:8 +0xf
```

Also adding some nil value checks for printing clusterCatalogs and clusterExtensions.

### DIFF
--- a/internal/cmd/internal/olmv1/catalog_create.go
+++ b/internal/cmd/internal/olmv1/catalog_create.go
@@ -13,7 +13,7 @@ import (
 
 // NewCatalogCreateCmd allows creating a new catalog
 func NewCatalogCreateCmd(cfg *action.Configuration) *cobra.Command {
-	i := v1action.NewCatalogCreate(cfg.Client)
+	i := v1action.NewCatalogCreate(cfg)
 	i.Logf = log.Printf
 
 	cmd := &cobra.Command{
--- a/internal/cmd/internal/olmv1/printing.go
+++ b/internal/cmd/internal/olmv1/printing.go
@@ -21,11 +21,16 @@ func printFormattedOperators(extensions ...olmv1.ClusterExtension) {
 
 	sortOperators(extensions)
 	for _, ext := range extensions {
+		var bundleName, bundleVersion string
+		if ext.Status.Install != nil {
+			bundleName = ext.Status.Install.Bundle.Name
+			bundleVersion = ext.Status.Install.Bundle.Version
+		}
 		age := time.Since(ext.CreationTimestamp.Time)
 		_, _ = fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
 			ext.Name,
-			ext.Status.Install.Bundle.Name,
-			ext.Status.Install.Bundle.Version,
+			bundleName,
+			bundleVersion,
 			ext.Spec.Source.SourceType,
 			status(ext.Status.Conditions, olmv1.TypeInstalled),
 			status(ext.Status.Conditions, olmv1.TypeProgressing),
@@ -41,13 +46,16 @@ func printFormattedCatalogs(catalogs ...olmv1.ClusterCatalog) {
 
 	sortCatalogs(catalogs)
 	for _, cat := range catalogs {
+		var lastUnpacked string
+		if cat.Status.LastUnpacked != nil {
+			duration.HumanDuration(time.Since(cat.Status.LastUnpacked.Time))
+		}
 		age := time.Since(cat.CreationTimestamp.Time)
-		lastUnpacked := time.Since(cat.Status.LastUnpacked.Time)
 		_, _ = fmt.Fprintf(tw, "%s\t%s\t%d\t%s\t%s\t%s\n",
 			cat.Name,
 			string(cat.Spec.AvailabilityMode),
 			cat.Spec.Priority,
-			duration.HumanDuration(lastUnpacked),
+			lastUnpacked,
 			status(cat.Status.Conditions, olmv1.TypeServing),
 			duration.HumanDuration(age),
 		)
@@ -59,6 +67,9 @@ func printFormattedCatalogs(catalogs ...olmv1.ClusterCatalog) {
 // name (asc), version (desc)
 func sortOperators(extensions []olmv1.ClusterExtension) {
 	slices.SortFunc(extensions, func(a, b olmv1.ClusterExtension) int {
+		if a.Status.Install == nil || b.Status.Install == nil {
+			return cmp.Compare(a.Name, b.Name)
+		}
 		return cmp.Or(
 			cmp.Compare(a.Name, b.Name),
 			-semver.MustParse(a.Status.Install.Bundle.Version).Compare(semver.MustParse(b.Status.Install.Bundle.Version)),
--- a/internal/pkg/v1/action/action_suite_test.go
+++ b/internal/pkg/v1/action/action_suite_test.go
@@ -12,8 +12,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
 	olmv1 "github.com/operator-framework/operator-controller/api/v1"
+
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
+
+const (
+	verbCreate = "create"
 )
 
 func TestCommand(t *testing.T) {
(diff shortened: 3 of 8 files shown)


internal/cmd/internal/olmv1/catalog_create.go
```diff
@@ -13,7 +14,10 @@ import (
 
 // NewCatalogCreateCmd allows creating a new catalog
 func NewCatalogCreateCmd(cfg *action.Configuration) *cobra.Command {
-	i := v1action.NewCatalogCreate(cfg.Client)
+	fmt.Println("lmao", cfg)
+	cfg.Load()
```
LalatenduMohanty, line 18: The question is why https://github.com/operator-framework/kubectl-operator/blob/main/internal/cmd/root.go#L61 does not work for us?

LalatenduMohanty: I think It was working at some point and with some recent changes it stopped working.

LalatenduMohanty: I thin we are doing something not right in https://github.com/operator-framework/kubectl-operator/blob/main/internal/cmd/root.go#L61

ankitathomas:
hold till https://github.com/operator-framework/kubectl-operator/pull/225 merges

internal/cmd/internal/olmv1/catalog_create.go
ankitathomas: It should work now

internal/cmd/internal/olmv1/catalog_create.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // NewCatalogCreateCmd allows creating a new catalog
 func NewCatalogCreateCmd(cfg *action.Configuration) *cobra.Command {
-	i := v1action.NewCatalogCreate(cfg.Client)
+	i := v1action.NewCatalogCreate(cfg)
```
LalatenduMohanty, line 16: Good catch. I have gone through the code many times but did not see this.

azych:
@ankitathomas could you explain what the issue with the client is and how it can be reproduced?

Can cfg `*action.Configuration` somehow be nil when it's passed to `NewCatalogCreateCmd`, which would make attempting to pass `.Client` panic?

If not and if the `Client` can be nil, how do the changes in this PR fix the situation?
We're still not validating if the `i.config.Client` is nil before calling it, just like we didin't validate if `i.client` was nil before.

Can the `cfg.Client` be modified/replaced during `cmd.Execute`?

internal/pkg/v1/action/action_suite_test.go
```diff
@@ -7,46 +7,85 @@ import (
...
-
-func (mg *mockGetter) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
-	mg.getCalled++
-	return mg.getErr
+func (c *fakeClient) Initialize() error {
+	scheme, err := action.NewScheme()
+	if err != nil {
+		return err
+	}
+	clientBuilder := fake.NewClientBuilder().WithInterceptorFuncs(interceptor.Funcs{
```
azych, line 57: that's cool, knowing about `WithInterceptorFuncs` and `"sigs.k8s.io/controller-runtime/pkg/client/interceptor"` definitely makes writing unit tests easier ;)

internal/cmd/internal/olmv1/catalog_create.go
ankitathomas: I didn't realize it at first either, I stumbled on this when I checked the implementation for the other subcommands within kubectl-operator

ankitathomas:
> Can cfg *action.Configuration somehow be nil when it's passed to NewCatalogCreateCmd, which would make attempting to pass .Client panic?

so, what happened was that when `v1action.NewCatalogCreate(cfg.Client)` gets called within `[NewCatalogCreateCmd](https://github.com/operator-framework/kubectl-operator/blob/main/internal/pkg/v1/action/catalog_create.go#L32)`, we pass it into CatalogCreate{}. 
```
	return &CatalogCreate{
		client: client,
		Logf:   func(string, ...interface{}) {},
	}
```
This would be ok if we initialize the actionConfig client beforehand, but that happens in the [PersistentPreRunE](https://github.com/operator-framework/kubectl-operator/blob/main/internal/cmd/root.go#L40) for the root command. However, since the command registration happens before `PersistentPreRunE` gets called, we end up passing the value of `cfg.Client` (nil) rather than the pointer to cfg.Client. So the way around it was to pass the entire actionConfig so by the time we end up using the client, [CatalogCreate.cfg.client](https://github.com/operator-framework/kubectl-operator/blob/main/internal/pkg/v1/action/catalog_create.go#L41) isn't nil

internal/cmd/internal/olmv1/printing.go
```diff
@@ -21,11 +21,16 @@ func printFormattedOperators(extensions ...olmv1.ClusterExtension) {
...
+			bundleName = ext.Status.Install.Bundle.Name
+			bundleVersion = ext.Status.Install.Bundle.Version
+		}
 		age := time.Since(ext.CreationTimestamp.Time)
 		_, _ = fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
 			ext.Name,
-			ext.Status.Install.Bundle.Name,
-			ext.Status.Install.Bundle.Version,
+			bundleName,
+			bundleVersion,
```
LalatenduMohanty, line 33: :+1:

LalatenduMohanty (approved):
/lgtm

LalatenduMohanty:
> hold till https://github.com/operator-framework/kubectl-operator/pull/225 merges

LalatenduMohanty:
> hold till https://github.com/operator-framework/kubectl-operator/pull/225 merges

@ankitathomas Can you please clarify why we need to wait for #225?

azych (approved):

LalatenduMohanty:
@ankitathomas Looks like lint test is failing with below errors

```
Run make lint
(re)installing /home/runner/go/bin/golangci-lint-v1.63.4
/home/runner/go/bin/golangci-lint-v1.63.4 --timeout 3m run
internal/pkg/v1/action/catalog_create_test.go:43:24   errcheck  Error return value of `mockClient.Initialize` is not checked
internal/pkg/v1/action/catalog_create_test.go:125:24  errcheck  Error return value of `mockClient.Client.Get` is not checked
internal/pkg/v1/action/catalog_create.go:13:6         unused    type `createClient` is unused
internal/pkg/v1/action/interfaces.go:9:6              unused    type `creator` is unused
internal/pkg/v1/action/catalog_create.go:7:1          gci       File is not properly formatted
internal/pkg/v1/action/catalog_create.go:10:1         gci       File is not properly formatted
internal/pkg/v1/action/action_suite_test.go:10:1      gci       File is not properly formatted
make: *** [Makefile:61: lint] Error 1
Error: Process completed with exit code 2.
```

LalatenduMohanty (approved):
/lgtm

LalatenduMohanty:
@ankitathomas Looks like the code for fixing the empty client in action config is missing after the rebase.

internal/cmd/internal/olmv1/printing.go
```diff
@@ -59,6 +67,9 @@ func printFormattedCatalogs(catalogs ...olmv1.ClusterCatalog) {
 // name (asc), version (desc)
 func sortOperators(extensions []olmv1.ClusterExtension) {
 	slices.SortFunc(extensions, func(a, b olmv1.ClusterExtension) int {
+		if a.Status.Install == nil || b.Status.Install == nil {
+			return cmp.Compare(a.Name, b.Name)
+		}
```
LalatenduMohanty, line 72: This is not related to the `empty client in action config` issue. Can you please add some context in the commit message? The commit message will help others understand the issue when they look back at the issue using `git blame`.

internal/pkg/v1/action/catalog_create.go
```diff
@@ -7,16 +7,12 @@ import (
...
-	creator
-	deleter
-	getter
-}
+	"github.com/operator-framework/kubectl-operator/pkg/action"
+)
 
 type CatalogCreate struct {
-	client         createClient
+	config         *action.Configuration
```
LalatenduMohanty, line 15: We should change it to `Config` instead of `config` to make it consistent with other members of the struct.

LalatenduMohanty (changes requested):

internal/pkg/v1/action/catalog_create.go
LalatenduMohanty: You are right, this is fine.

LalatenduMohanty (approved):
/lgtm

